PR #6, opened 2023-05-24
Remove plugin artifacts and update readme to pull from releases

kate-goldenring:
Manifest and plugin tarball are now generated each merge.


karthik2804 (approved):
LGTM!

PR #7, opened 2023-05-25
Add `--variables` deploy flag and `variables` subcommand

kate-goldenring:
- Draft because i am not sure if we want to wait to merge this once we flip this feature on and whether the cloud plugin will come before config vars
- Moves most of the body of the DeployCommand's `run` function into a sharable `login_connection` function that will likely be used by most subcommands.
- Identifies an app by app manifest `name`

## Example
App with the following manifest:
```toml
spin_manifest_version = "1"
authors = ["Fermyon Engineering <engineering@fermyon.com>"]
description = "A simple application that returns hello."
name = "variables-plugin"
trigger = { type = "http", base = "/" }
version = "1.0.0"

[variables]
host = {required = true}

[[component]]
id = "hello"
source = "target/wasm32-wasi/release/http_rust.wasm"
description = "A simple component that returns hello."
[component.trigger]
route = "/hello"
[component.config]
host = "{{ host }}"
[component.build]
command = "cargo build --target wasm32-wasi --release"
```
and handle
```rust
#[http_component]
fn hello_world(req: Request) -> Result<Response> {
    let host = config::get("host")?;
    Ok(http::Response::builder()
        .status(200)
        .header("foo", "bar")
        .body(Some(format!("Hello, {host}\n").into()))?)
}
```

Can deploy it with a config variable and then update it. I'm using the [cloud plugin version of Spin from this PR](https://github.com/fermyon/spin/pull/1452).

```bash
cargo run -- deploy --variable host=wowzee  --from examples/http-rust 
# Hello, wowzee
cargo run -- cloud variables set host=yippee unused=value --app variables-plugin
# Hello, yippee
```

### DIFF
--- a/src/commands/variables.rs
+++ b/src/commands/variables.rs
@@ -1,5 +1,5 @@
 use anyhow::{anyhow, Context, Result};
-use clap::Parser;
+use clap::{Args, Parser};
 use cloud::client::{Client as CloudClient, ConnectionConfig};
 use spin_common::arg_parser::parse_kv;
 use uuid::Uuid;
@@ -11,12 +11,23 @@ use crate::{
 
 /// Manage Spin application variables
 #[derive(Parser, Debug)]
-#[clap(about = "Log into the Fermyon Platform")]
-pub struct VariablesCommand {
-    /// Variable pair to set
-    #[clap(name = VARIABLES_SET_OPT, short = 's', long = "set", parse(try_from_str = parse_kv))]
-    pub variables_to_set: Vec<(String, String)>,
+#[clap(about = "Manage Spin application variables")]
+pub enum VariablesCommand {
+    /// Set variable pairs
+    Set(SetCommand),
+}
 
+#[derive(Parser, Debug)]
+pub struct SetCommand {
+    /// Variable pair to set
+    #[clap(name = VARIABLES_SET_OPT, parse(try_from_str = parse_kv))]
+    pub variables_to_set: Vec<(String, String)>,
+    #[clap(flatten)]
+    common: CommonArgs,
+}
+
+#[derive(Debug, Default, Args)]
+struct CommonArgs {
     /// Deploy to the Fermyon instance saved under the specified name.
     /// If omitted, Spin deploys to the default unnamed instance.
     #[clap(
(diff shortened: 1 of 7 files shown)


radu-matei (approved):
Tested the flow to deploy an application and set pair, and the separate flow to set a variable pair, and everything worked great for me!

The only structural question is whether `cloud variable set` should be a separate subcommand rather than the current experience, in which it is a flag. I imagine having to add deletion and list subcommands means we will make this a bit more clear.

The end to end flow is working great for me!

src/commands/variables.rs
```diff
@@ -0,0 +1,66 @@
...
+
+use crate::{
+    commands::deploy::{get_app_id_cloud, login_connection},
+    opts::*,
+};
+
+/// Manage Spin application variables
+#[derive(Parser, Debug)]
+#[clap(about = "Log into the Fermyon Platform")]
+pub struct VariablesCommand {
```
fibonacci1729, line 15: Could we add a `vars` alias?

src/commands/variables.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// Variable pair to set
+    #[clap(name = VARIABLES_SET_OPT, short = 's', long = "set", parse(try_from_str = parse_kv))]
+    pub variables_to_set: Vec<(String, String)>,
```
fibonacci1729, line 18: I think we should make `set` a subcommand here.

kate-goldenring: definitely. will do!

src/commands/variables.rs
```diff
@@ -0,0 +1,91 @@
...
+    pub app: String,
+}
+
+/// Information needed to connect to Fermyon Cloud to manage a specific Spin application
+struct AppManagmentInfo {
+    app_id: Uuid,
+    client: CloudClient,
+}
+
+impl AppManagmentInfo {
```
kate-goldenring, line 50: I anticipate this being shared across all subcommands in the future so i moved it into a separate struct but it might be an antipattern for a struct that has no methods. Should this be a type instead @fibonacci1729 ? I do like having the named fields instead of positional (`info.0`).

src/commands/variables.rs
```diff
@@ -0,0 +1,91 @@
...
+};
+
+/// Manage Spin application variables
+#[derive(Parser, Debug)]
+#[clap(about = "Manage Spin application variables")]
+pub enum VariablesCommand {
+    Set(SetCommand),
+}
+
+#[derive(Parser, Debug)]
```
fibonacci1729, line 19: Might need a doc comment here so it displays in the help.

fibonacci1729 (approved):
One minor thing re: doc comment on SetCommand, great work! LGTM

PR #16, opened 2023-06-04
Add Variables subcommands for listing and deleting variables

kate-goldenring:
Adds `list` and `delete` subcommands to the `variables` CLI

Also, hides the `variables` subcommand and `--variable` deploy flag. Can unhide when we release the feature

### DIFF
--- a/src/commands/variables.rs
+++ b/src/commands/variables.rs
@@ -114,12 +114,8 @@ impl VariablesCommand {
                     .map(|var| from_str(var))
                     .collect::<Result<Vec<Variable>, _>>()
                     .context("could not parse variable")?;
-                if !var_names.is_empty() {
-                    println!();
-                    println!("App {} has the following variables:", cmd.common.app)
-                }
                 for v in var_names {
-                    println!("    {}", v.key);
+                    println!("{}", v.key);
                 }
             }
         }
--- a/src/commands/variables.rs
+++ b/src/commands/variables.rs
@@ -1,6 +1,8 @@
 use anyhow::{anyhow, Context, Result};
 use clap::{Args, Parser};
 use cloud::client::{Client as CloudClient, ConnectionConfig};
+use serde::Deserialize;
+use serde_json::from_str;
 use spin_common::arg_parser::parse_kv;
 use uuid::Uuid;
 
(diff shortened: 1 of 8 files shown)


src/commands/variables.rs
```diff
@@ -73,6 +98,30 @@ impl VariablesCommand {
...
+            Self::List(cmd) => {
+                let info =
+                    AppManagmentInfo::new(cmd.common.deployment_env_id.as_deref(), &cmd.common.app)
+                        .await?;
+                let vars = get_variables(&info.client, info.app_id).await?;
+                let var_names = vars
+                    .iter()
+                    .map(|var| from_str(var))
+                    .collect::<Result<Vec<Variable>, _>>()
+                    .context("could not parse variable")?;
```
fibonacci1729, line 116: For consistency with other contexts here:

```suggestion
                    .context("Could not parse variable")?;
```

fibonacci1729 (approved):

src/commands/variables.rs
```diff
@@ -73,6 +98,30 @@ impl VariablesCommand {
...
+                if !var_names.is_empty() {
+                    println!();
+                    println!("App {} has the following variables:", cmd.common.app)
+                }
+                for v in var_names {
```
kate-goldenring, line 121: @itowlson I think i could use help on making this a prettier experience than tabbed prints. Or maybe this is sufficient

itowlson: @kate-goldenring For multi-value data we've used tables (e.g. `spin templates list`), but there doesn't seem much point when we're just listing names (cf. `spin plugins list`).

That said, I'm not sure there's value in the header and tabs.  How would it feel to just have:

```
$ spin cloud variables list
alpha
bravo
charlie
```

If we changed it to show extra info like the value or if the variable was secret or was modified from its default value or whatever, then `comfy-table` has been our friend.

kate-goldenring: Sounds good to me. Simple and clean. Fixed it

src/commands/variables.rs
```diff
@@ -73,6 +98,26 @@ impl VariablesCommand {
...
+                    AppManagmentInfo::new(cmd.common.deployment_env_id.as_deref(), &cmd.common.app)
+                        .await?;
+                let vars = get_variables(&info.client, info.app_id).await?;
+                let var_names = vars
+                    .iter()
+                    .map(|var| from_str(var))
+                    .collect::<Result<Vec<Variable>, _>>()
+                    .context("could not parse variable")?;
+                for v in var_names {
+                    println!("{}", v.key);
```
radu-matei, line 118: Not blocking this PR, but do we want to format a table head? For example, the output for the templates list command:
```
+------------------------------------------------------------------------+
| Name                Description                                        |
+========================================================================+

```

radu-matei (approved):

PR #19, opened 2023-06-06
Update deploy to use OCI as distribution mechanism

radu-matei:


src/commands/deploy.rs
```diff
@@ -393,48 +397,50 @@ impl DeployCommand {
...
+                    + "-"
+                    + &buildinfo
+            }
+            None => cloud_registry_url.domain().unwrap().to_owned() + "/" + &application.info.name,
         };
 
-        let bindle_id = spin_bindle::prepare_bindle(&self.app()?, buildinfo, dest_dir)
-            .await
-            .map_err(crate::wrap_prepare_bindle_error)?;
+        println!("Uploading {} to Fermyon Cloud...", reference);
```
bacongobbler, line 429: Would you mind removing the registry URL here? I don't think there's any action we want users to take with this URL. e.g. they can't use `spin registry push` or `spin registry pull` with this URL IIRC.

radu-matei: That's a good point. I wasn't sure it should be there to begin with, so I'll remove it.

vdice:
Closing in favor of https://github.com/fermyon/cloud-plugin/pull/57, updated with the latest changes from `main`.

src/commands/deploy.rs
vdice: This has been updated in https://github.com/fermyon/cloud-plugin/pull/57/commits/dd19eec9cc4d98147e2fab6121065101923188e7

PR #37, opened 2023-06-23
Validate application variables are set before deploy

### LINKED ISSUE #35: Validate required variables are defined at deploy time
This was flagged to me by @tschneidereit.  It appears that, when you deploy an application, if any `required` variable is not either set during deployment or already set in the Cloud, the deployment goes ahead anyway, and then fails at runtime when the application requests the required variable.

A possible solution is to ask the Cloud at deploy time what variables are already defined, combine that with the list of variables provided in the deploy command, and cross-check that against the application's `required` variables.  If any are missing, we could prompt for values or fail the deployment.

(There is a possible race condition with doing it on the client, if a user sneakily deletes a variable from the Cloud between the "list variables" query and the deployment, but that's not a big concern.)

itowlson:
Fixes #35.

At the moment, this _blocks_ deployment if any required variables are not defined.  This could be reduced to a warning, or we could drop into an interactive session that prompts for the missing variable values - reviewers please share what you think is right!

Because variables may have been defined during a previous deployment, the validator checks what variables are already defined in the cloud.  If an error occurs during this check, the validator currently adopts the principle of cheery optimism: it lets the user know that it couldn't check, but allows the deployment to continue.  Again, reviewers' opinions on whether this is the right thing would be welcome!

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -25,6 +25,7 @@ use tokio::fs;
 use tracing::instrument;
 
 use std::{
+    collections::HashSet,
     fs::File,
     io::{self, copy, Write},
     path::PathBuf,
@@ -32,7 +33,7 @@ use std::{
 use url::Url;
 use uuid::Uuid;
 
-use crate::commands::variables::set_variables;
+use crate::commands::variables::{get_variables, set_variables, Variable};
 
 use crate::{
     commands::login::{LoginCommand, LoginConnection},
@@ -145,6 +146,7 @@ impl DeployCommand {
         let cfg = cfg_any.into_v1();
 
         validate_cloud_app(&cfg)?;
+        self.validate_deployment_environment(&cfg, &client).await?;
 
         match cfg.info.trigger {
             ApplicationTrigger::Http(_) => {}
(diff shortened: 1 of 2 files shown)


src/commands/variables.rs
```diff
@@ -12,8 +12,8 @@ use crate::{
 };
 
 #[derive(Deserialize)]
-struct Variable {
-    key: String,
+pub(crate) struct Variable {
+    pub(crate) key: String,
```
fibonacci1729, line 16: This could just be `pub`

src/commands/deploy.rs
```diff
@@ -322,6 +324,73 @@ impl DeployCommand {
...
+        let unprovided_variables = unprovided_variables
+            .difference(&extant_variables)
+            .map(|v| v.as_str())
+            .collect::<Vec<_>>();
+        if unprovided_variables.is_empty() {
+            return Ok(());
+        }
+
+        let list_text = unprovided_variables.join(", ");
+        Err(anyhow!("The application requires values for the following variable(s) which have not been set: {list_text}. Use the --variables flag to provide values."))
```
fibonacci1729, line 391: I believe the flag is `--variable`

itowlson: Fie!  You believe right - thanks for the spot

fibonacci1729:
Just an idea (definirely not blocking this PR) but should we consider applying this eager guarantee approximation for spin up as well? I don't have strong feelings about this ATM just wanting to note that for local dev missing config will still be a runtime error.

itowlson:
I think we've discussed it / looked at it before, and it faltered on needing to interrogate sources that were only defined at runtime (particularly since understanding of that stuff is done on the trigger side rather than the `up` side).  It could be done for sure though, just a matter of knuckling down to it!

kate-goldenring (approved):
LGTM! Thanks @itowlson!

PR #47, opened 2023-07-21
Add a header to the databases list output

rylev:
This adds a header to the list of running `cloud-plugin sqlite list`. This is particularly helpful when there are no databases as it otherwise just prints nothing.

image


kate-goldenring (approved):

itowlson:
nit: In the "no databases" case it might be nicer to print "There are no databases" or "You have no databases set up" or something like that rather than "Databases (0)" - we can always refine this later though

PR #49, opened 2023-07-21
Add execute sql boiler impl

rylev:
Adds the boiler plate code for calling (as of yet non-existent) endpoint in nimbus for executing SQL.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -442,7 +442,7 @@ dependencies = [
 [[package]]
 name = "cloud-openapi"
 version = "0.1.0"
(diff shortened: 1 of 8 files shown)


kate-goldenring:
@rylev as the sqld-proxy service is written and as `Lookup` and gluon are implemented, `app_id` is required for execute. I think in order to add this command, we need to update `Lookup` in gluon to allow for just `project_id` and `name` for a lookup. Then we can run `spin cloud sqlite execute "BLAH"`. Otherwise, it is really hard right now for a user to know what db is associated with what app.

rylev:
Ok this is ready. We no longer need to send app id so @kate-goldenring's concern above is addressed.

PR #58, opened 2023-08-11
add cloud logs command

rajatjindal:
```
spin-cloud-logs 0.1.2 (b977558 2023-08-11)
Logs fetches app logs from Fermyon Cloud

USAGE:
    spin cloud logs [OPTIONS] --app-name <app-name>

OPTIONS:
        --app-name <app-name>
            App name

        --environment-name <environment-name>
            Find app to fetch logs for in the Fermyon instance saved under the specified name. If
            omitted, Spin looks for app in default unnamed instance [env:
            FERMYON_DEPLOYMENT_ENVIRONMENT=]

        --follow
            Follow logs output

    -h, --help
            Print help information

        --interval <interval>
            interval in secs to refresh logs from cloud [default: 5]

        --tail <tail>
            Number of lines to show from the end of the logs [default: 10]

    -V, --version
            Print version information

```

### DIFF
--- a/src/commands/logs.rs
+++ b/src/commands/logs.rs
@@ -144,17 +144,13 @@ fn print_logs(entries: &[Entry], show_timestamp: bool) -> Option<&str> {
                 continue;
             };
 
-            match &log_entry.time {
-                Some(time) => {
-                    if show_timestamp {
-                        println!("[{time}] {log}");
-                    } else {
-                        println!("{log}");
-                    }
-
-                    since = Some(time.as_str());
+            if let Some(time) = &log_entry.time {
+                if show_timestamp {
+                    println!("[{time}] {log}");
+                } else {
+                    println!("{log}");
                 }
-                None => (),
+                since = Some(time.as_str());
             }
         }
     }
--- a/src/commands/logs.rs
+++ b/src/commands/logs.rs
@@ -1,0 +1,192 @@
+use std::ops::Sub;
+use std::time::Duration;
+
+use anyhow::{bail, Context, Result};
+use chrono::Utc;
+use cloud::{CloudClientExt, CloudClientInterface};
+use cloud_openapi::models::Entry;
+use std::option::Option;
+
+use crate::commands::create_cloud_client;
+use crate::opts::*;
+use clap::Parser;
+use uuid::Uuid;
+
+/// fetch logs for an app from Fermyon Cloud
+#[derive(Parser, Debug)]
+pub struct LogsCommand {
+    /// Use the Fermyon instance saved under the specified name.
+    /// If omitted, Spin looks for app in default unnamed instance.
+    #[clap(
+        name = "environment-name",
+        long = "environment-name",
+        env = DEPLOYMENT_ENV_NAME_ENV
+    )]
+    pub deployment_env_id: Option<String>,
+
+    /// App name
+    pub app: String,
+
+    /// Follow logs output
+    #[clap(name = "follow", long = "follow")]
+    pub follow: bool,
+
+    /// Number of lines to show from the end of the logs
+    #[clap(name = "tail", long = "tail", default_value = "10")]
+    pub max_lines: i32,
+
+    /// Interval in seconds to refresh logs from cloud
+    #[clap(parse(try_from_str = parse_interval), name="interval", long="interval", default_value = "2")]
+    pub interval_secs: std::time::Duration,
+
+    /// Only return logs newer than a relative duration. The duration format is a number
+    /// and a unit, where the unit is 's' for seconds, 'm' for minutes, 'h' for hours
+    /// or 'd' for days (e.g. "30m" for 30 minutes ago).  The default it 7 days.
+    #[clap(parse(try_from_str = parse_duration), name="since", long="since", default_value = "7d")]
+    pub since: std::time::Duration,
+
+    /// Show timestamps
+    #[clap(
+        name = "show-timestamps",
+        long = "show-timestamps",
+        default_value = "true",
+        action = clap::ArgAction::Set
+    )]
+    pub show_timestamp: bool,
+}
+
+impl LogsCommand {
+    pub async fn run(self) -> Result<()> {
+        let client = create_cloud_client(self.deployment_env_id.as_deref()).await?;
+        self.logs(&client).await
+    }
+
+    async fn logs(self, client: &impl CloudClientInterface) -> Result<()> {
+        let app_id = client
+            .get_app_id(&self.app)
+            .await
+            .with_context(|| format!("failed to find app with name {:?}", &self.app))?
+            .with_context(|| format!("app with name {:?} not found", &self.app))?;
+
+        fetch_logs_and_print_loop(
+            client,
+            app_id,
+            self.follow,
+            self.interval_secs,
+            self.max_lines,
+            self.since,
+            self.show_timestamp,
+        )
+        .await?;
+
+        Ok(())
+    }
+}
+
+async fn fetch_logs_and_print_loop(
+    client: &impl CloudClientInterface,
+    app_id: Uuid,
+    follow: bool,
+    interval: Duration,
+    max_lines: i32,
+    since: Duration,
+    show_timestamp: bool,
+) -> Result<()> {
+    let mut curr_since = Utc::now().sub(since).to_rfc3339();
+    curr_since =
+        fetch_logs_and_print_once(client, app_id, Some(max_lines), curr_since, show_timestamp)
+            .await?;
+
+    if !follow {
+        return Ok(());
+    }
+
+    loop {
+        tokio::time::sleep(interval).await;
+        curr_since =
+            fetch_logs_and_print_once(client, app_id, None, curr_since, show_timestamp).await?;
+    }
+}
+
+async fn fetch_logs_and_print_once(
+    client: &impl CloudClientInterface,
+    app_id: Uuid,
+    max_lines: Option<i32>,
+    since: String,
+    show_timestamp: bool,
+) -> Result<String> {
+    let entries = client
+        .app_logs_raw(app_id.to_string(), max_lines, Some(since.to_string()))
+        .await?
+        .entries;
+
+    if entries.is_empty() {
+        return Ok(since.to_owned());
+    }
+
+    let updated_since = print_logs(&entries, show_timestamp);
+    if let Some(u) = updated_since {
+        return Ok(u.to_owned());
+    }
+
+    Ok(since)
+}
+
+fn print_logs(entries: &[Entry], show_timestamp: bool) -> Option<&str> {
+    let mut since = None;
+    for entry in entries.iter().rev() {
+        let Some(log_lines) = entry.log_lines.as_ref() else {
+            continue;
+        };
+
+        for log_entry in log_lines {
+            let Some(log) = log_entry.line.as_ref() else {
+                continue;
+            };
+
+            match &log_entry.time {
+                Some(time) => {
+                    if show_timestamp {
+                        println!("[{time}] {log}");
+                    } else {
+                        println!("{log}");
+                    }
+
+                    since = Some(time.as_str());
+                }
+                None => (),
+            }
+        }
+    }
+
+    since
+}
+
+fn parse_duration(arg: &str) -> anyhow::Result<std::time::Duration> {
+    let duration = if let Some(parg) = arg.strip_suffix('s') {
+        let value = parg.parse()?;
+        std::time::Duration::from_secs(value)
+    } else if let Some(parg) = arg.strip_suffix('m') {
+        let value: u64 = parg.parse()?;
+        std::time::Duration::from_secs(value * 60)
+    } else if let Some(parg) = arg.strip_suffix('h') {
+        let value: u64 = parg.parse()?;
+        std::time::Duration::from_secs(value * 60 * 60)
+    } else if let Some(parg) = arg.strip_suffix('d') {
+        let value: u64 = parg.parse()?;
+        std::time::Duration::from_secs(value * 24 * 60 * 60)
+    } else {
+        bail!(r#"since must be a number followed by an allowed unit ("300s", "5m", "4h" or "1d")"#);
+    };
+
+    Ok(duration)
+}
+
+fn parse_interval(arg: &str) -> anyhow::Result<std::time::Duration> {
+    let value = arg.parse()?;
+    if value < 2 {
+        bail!("interval cannot be less than 2 seconds")
+    }
+
+    Ok(std::time::Duration::from_secs(value))
+}
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -2,6 +2,7 @@ pub mod apps;
 pub mod deploy;
 pub mod link;
 pub mod login;
+pub mod logs;
 pub mod sqlite;
 pub mod variables;
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,6 +10,7 @@ use commands::{
     deploy::DeployCommand,
     link::{LinkCommand, UnlinkCommand},
     login::LoginCommand,
+    logs::LogsCommand,
     sqlite::SqliteCommand,
     variables::VariablesCommand,
 };
@@ -35,6 +36,8 @@ enum CloudCli {
     Deploy(DeployCommand),
     /// Login to Fermyon Cloud
     Login(LoginCommand),
+    /// Fetch logs for an app from Fermyon Cloud
+    Logs(LogsCommand),
     /// Manage Spin application variables
     #[clap(subcommand, alias = "vars")]
     Variables(VariablesCommand),
@@ -61,6 +64,7 @@ async fn main() -> Result<(), Error> {
         CloudCli::Apps(cmd) => cmd.run().await,
         CloudCli::Deploy(cmd) => cmd.run().await,
         CloudCli::Login(cmd) => cmd.run().await,
+        CloudCli::Logs(cmd) => cmd.run().await,
         CloudCli::Variables(cmd) => cmd.run().await,
         CloudCli::Sqlite(cmd) => cmd.run().await,
         CloudCli::Link(cmd) => cmd.run().await,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,7 +105,7 @@ checksum = "f1b6f5d7df27bd294849f8eec66ecfc63d11814df7a4f5d74168a2394467b776"
 dependencies = [
  "async-channel",
  "async-executor",
- "async-io",
+ "async-io 1.13.0",
  "async-lock",
  "blocking",
  "futures-lite",
@@ -125,11 +125,31 @@ dependencies = [
  "futures-lite",
  "log",
  "parking",
- "polling",
- "rustix 0.37.26",
+ "polling 2.8.0",
+ "rustix 0.37.27",
  "slab",
  "socket2 0.4.10",
  "waker-fn",
+]
+
+[[package]]
+name = "async-io"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "10da8f3146014722c89e7859e1d7bb97873125d7346d10ca642ffab794355828"
+dependencies = [
+ "async-lock",
+ "cfg-if",
+ "concurrent-queue",
+ "futures-io",
+ "futures-lite",
+ "parking",
+ "polling 3.3.0",
+ "rustix 0.38.21",
+ "slab",
+ "tracing",
+ "waker-fn",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -147,30 +167,30 @@ version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ea6438ba0a08d81529c69b36700fa2f95837bfe3e776ab39cde9c14d9149da88"
 dependencies = [
- "async-io",
+ "async-io 1.13.0",
  "async-lock",
  "async-signal",
  "blocking",
  "cfg-if",
- "event-listener 3.0.0",
+ "event-listener 3.0.1",
  "futures-lite",
- "rustix 0.38.20",
+ "rustix 0.38.21",
  "windows-sys 0.48.0",
 ]
 
 [[package]]
 name = "async-signal"
-version = "0.2.4"
+version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d2a5415b7abcdc9cd7d63d6badba5288b2ca017e3fbd4173b8f405449f1a2399"
+checksum = "9e47d90f65a225c4527103a8d747001fc56e375203592b25ad103e1ca13124c5"
 dependencies = [
- "async-io",
+ "async-io 2.1.0",
  "async-lock",
  "atomic-waker",
  "cfg-if",
  "futures-core",
  "futures-io",
- "rustix 0.38.20",
+ "rustix 0.38.21",
  "signal-hook-registry",
  "slab",
  "windows-sys 0.48.0",
@@ -184,7 +204,7 @@ checksum = "62565bb4402e926b29953c785397c6dc0391b7b446e45008b0049eb43cec6f5d"
 dependencies = [
  "async-channel",
  "async-global-executor",
- "async-io",
+ "async-io 1.13.0",
  "async-lock",
  "async-process",
  "crossbeam-utils",
@@ -452,7 +472,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
- "tokio-util 0.7.9",
+ "tokio-util 0.7.10",
  "tracing",
  "uuid",
 ]
@@ -460,7 +480,7 @@ dependencies = [
 [[package]]
 name = "cloud-openapi"
 version = "0.1.0"
-source = "git+https://github.com/fermyon/cloud-openapi?rev=3e70369e3fc9574e262827332157da40da0a4f66#3e70369e3fc9574e262827332157da40da0a4f66"
+source = "git+https://github.com/fermyon/cloud-openapi?rev=ce1e916110b9a9e59a1171ac364f0b6e23908428#ce1e916110b9a9e59a1171ac364f0b6e23908428"
 dependencies = [
  "reqwest",
  "serde",
@@ -562,9 +582,9 @@ checksum = "e496a50fda8aacccc86d7529e2c1e0892dbd0f898a6b5645b5561b89c3210efa"
 
 [[package]]
 name = "cpufeatures"
-version = "0.2.10"
+version = "0.2.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3fbc60abd742b35f2492f808e1abbb83d45f72db402e14c55057edc9c7b1e9e4"
+checksum = "ce420fe07aecd3e67c5f910618fe65e94158f6dcc0adf44e00d69ce2bdfe0fd0"
 dependencies = [
  "libc",
 ]
@@ -830,9 +850,9 @@ checksum = "0206175f82b8d6bf6652ff7d71a1e27fd2e4efde587fd368662814d6ec1d9ce0"
 
 [[package]]
 name = "event-listener"
-version = "3.0.0"
+version = "3.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "29e56284f00d94c1bc7fd3c77027b4623c88c1f53d8d2394c6199f2921dea325"
+checksum = "01cec0252c2afff729ee6f00e903d479fba81784c8e2bd77447673471fdfaea1"
 dependencies = [
  "concurrent-queue",
  "parking",
@@ -923,9 +943,9 @@ checksum = "6c2141d6d6c8512188a7891b4b01590a45f6dac67afb4f255c4124dbb86d4eaa"
 
 [[package]]
 name = "futures"
-version = "0.3.28"
+version = "0.3.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "23342abe12aba583913b2e62f22225ff9c950774065e4bfb61a19cd9770fec40"
+checksum = "da0290714b38af9b4a7b094b8a37086d1b4e61f2df9122c3cad2577669145335"
 dependencies = [
  "futures-channel",
  "futures-core",
@@ -938,9 +958,9 @@ dependencies = [
 
 [[package]]
 name = "futures-channel"
-version = "0.3.28"
+version = "0.3.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "955518d47e09b25bbebc7a18df10b81f0c766eaf4c4f1cccef2fca5f2a4fb5f2"
+checksum = "ff4dd66668b557604244583e3e1e1eada8c5c2e96a6d0d6653ede395b78bbacb"
 dependencies = [
  "futures-core",
  "futures-sink",
@@ -948,15 +968,15 @@ dependencies = [
 
 [[package]]
 name = "futures-core"
-version = "0.3.28"
+version = "0.3.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4bca583b7e26f571124fe5b7561d49cb2868d79116cfa0eefce955557c6fee8c"
+checksum = "eb1d22c66e66d9d72e1758f0bd7d4fd0bee04cad842ee34587d68c07e45d088c"
 
 [[package]]
 name = "futures-executor"
-version = "0.3.28"
+version = "0.3.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ccecee823288125bd88b4d7f565c9e58e41858e47ab72e8ea2d64e93624386e0"
+checksum = "0f4fb8693db0cf099eadcca0efe2a5a22e4550f98ed16aba6c48700da29597bc"
 dependencies = [
  "futures-core",
  "futures-task",
@@ -965,9 +985,9 @@ dependencies = [
 
 [[package]]
 name = "futures-io"
-version = "0.3.28"
+version = "0.3.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4fff74096e71ed47f8e023204cfd0aa1289cd54ae5430a9523be060cdb849964"
+checksum = "8bf34a163b5c4c52d0478a4d757da8fb65cabef42ba90515efee0f6f9fa45aaa"
 
 [[package]]
 name = "futures-lite"
@@ -986,9 +1006,9 @@ dependencies = [
 
 [[package]]
 name = "futures-macro"
-version = "0.3.28"
+version = "0.3.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "89ca545a94061b6365f2c7355b4b32bd20df3ff95f02da9329b34ccc3bd6ee72"
+checksum = "53b153fd91e4b0147f4aced87be237c98248656bb01050b96bf3ee89220a8ddb"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -997,21 +1017,21 @@ dependencies = [
 
 [[package]]
 name = "futures-sink"
-version = "0.3.28"
+version = "0.3.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f43be4fe21a13b9781a69afa4985b0f6ee0e1afab2c6f454a8cf30e2b2237b6e"
+checksum = "e36d3378ee38c2a36ad710c5d30c2911d752cb941c00c72dbabfb786a7970817"
 
 [[package]]
 name = "futures-task"
-version = "0.3.28"
+version = "0.3.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "76d3d132be6c0e6aa1534069c705a74a5997a356c0dc2f86a47765e5617c5b65"
+checksum = "efd193069b0ddadc69c46389b740bbccdd97203899b48d09c5f7969591d6bae2"
 
 [[package]]
 name = "futures-util"
-version = "0.3.28"
+version = "0.3.29"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "26b01e40b772d54cf6c6d721c1d1abd0647a0106a12ecaa1c186273392a69533"
+checksum = "a19526d624e703a3179b3d322efec918b6246ea0fa51d41124525f00f1cc8104"
 dependencies = [
  "futures-channel",
  "futures-core",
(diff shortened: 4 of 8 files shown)


src/commands/logs.rs
```diff
@@ -0,0 +1,120 @@
...
+impl LogsCommand {
+    pub async fn run(self) -> Result<()> {
+        let login_connection = login_connection(self.deployment_env_id.as_deref()).await?;
+        let app_name: String = self.app_name.clone();
+        self.tail_logs(login_connection, app_name)
+            .await
+            .map_err(|e| anyhow!("{:?}\n\nLearn more at {}", e, "DEVELOPER_CLOUD_FAQ"))
+    }
+
+    async fn tail_logs(self, login_connection: LoginConnection, app_name: String) -> Result<()> {
```
rylev, line 50: Nit:
```suggestion
    async fn tail_logs(self, login_connection: LoginConnection, app_name: &str) -> Result<()> {
```

src/commands/logs.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    /// interval in secs to refresh logs from cloud
+    #[clap(name = "interval", long = "interval", default_value = "5")]
+    pub interval_secs: u64,
+}
+
+impl LogsCommand {
+    pub async fn run(self) -> Result<()> {
+        let login_connection = login_connection(self.deployment_env_id.as_deref()).await?;
+        let app_name: String = self.app_name.clone();
```
rylev, line 44: The change below means that you don't need to clone.
```suggestion
```

src/commands/logs.rs
```diff
@@ -0,0 +1,120 @@
...
+
+async fn fetch_and_print_latest_logs(
+    client: &CloudClient,
+    channel_id: Uuid,
+    printed_so_far: usize,
+) -> Result<usize> {
+    let logs_resp = client.channel_logs(channel_id.to_string()).await?;
+    let count = logs_resp.logs.len();
+    print_lastn_logs(logs_resp.logs, count - printed_so_far);
+    Ok(count)
```
rylev, line 113:
```suggestion
    fetch_and_print_lastn_logs(client, channel_id, count- printed_so_far).await
```

src/commands/logs.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+fn print_lastn_logs(logs: Vec<String>, n: usize) {
```
rylev, line 116: Nit: it's usually best practice to not take ownership if you don't need it. 
```suggestion
fn print_lastn_logs(logs: &[String], n: usize) {
```

src/commands/mod.rs
```diff
@@ -26,3 +27,33 @@ pub(crate) async fn get_app_id_cloud(cloud_client: &CloudClient, name: &str) ->
         None => bail!("No app with name: {}", name),
     }
 }
+
+async fn get_channel_id_cloud(
```
rylev, line 31: Should helpers like this live in the cloud crate?

src/commands/logs.rs
```diff
@@ -0,0 +1,120 @@
...
+        let connection_config = ConnectionConfig {
+            url: login_connection.url.to_string(),
+            insecure: login_connection.danger_accept_invalid_certs,
+            token: login_connection.token.clone(),
+        };
+
+        let client = CloudClient::new(connection_config.clone());
+        let app_id = get_app_id_cloud(&client, app_name.as_str()).await?;
+        let channel_id =
+            get_channel_id_cloud(&client, SPIN_DEPLOY_CHANNEL_NAME.to_string(), app_id).await?;
```
rylev, line 60: The error for this talks about a channel id which the user is not going to know anything about. Do we want to use `context` or `map_err` to provide a better error message?

rylev (changes requested):
I'd like to figure out if there's a way to do this without repeatedly downloading *all* logs.

src/commands/logs.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Number of lines to show from the end of the logs
+    #[clap(name = "tail", long = "tail", default_value = "10")]
+    pub lastn: usize,
+
+    /// interval in secs to refresh logs from cloud
+    #[clap(name = "interval", long = "interval", default_value = "5")]
+    pub interval_secs: u64,
+}
+
+impl LogsCommand {
```
kate-goldenring, line 41:
```suggestion
impl LogsCommand {
```

src/commands/logs.rs
```diff
@@ -0,0 +1,120 @@
+use anyhow::{anyhow, Result};
+
+use crate::commands::deploy::{login_connection, SPIN_DEPLOY_CHANNEL_NAME};
+use crate::commands::login::LoginConnection;
+use crate::commands::{get_app_id_cloud, get_channel_id_cloud};
```
kate-goldenring, line 5: Can use helper for creating the client. See below for continued suggestion
```suggestion
use crate::commands::{get_app_id_cloud, get_channel_id_cloud, create_cloud_client};
```

src/commands/logs.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    async fn tail_logs(self, login_connection: LoginConnection, app_name: String) -> Result<()> {
+        let connection_config = ConnectionConfig {
+            url: login_connection.url.to_string(),
+            insecure: login_connection.danger_accept_invalid_certs,
+            token: login_connection.token.clone(),
+        };
+
+        let client = CloudClient::new(connection_config.clone());
+        let app_id = get_app_id_cloud(&client, app_name.as_str()).await?;
```
kate-goldenring, line 58: If you create the client in `run` using `create_cloud_client`, you can pass it and simplify this
```suggestion
    async fn tail_logs(self, client: &CloudClient, app_name: String) -> Result<()> {
        let app_id = get_app_id_cloud(client, app_name.as_str()).await?;
```

src/commands/logs.rs
```diff
@@ -0,0 +1,120 @@
...
+    let count = logs_resp.logs.len();
+    print_lastn_logs(logs_resp.logs, count - printed_so_far);
+    Ok(count)
+}
+
+fn print_lastn_logs(logs: Vec<String>, n: usize) {
+    for log in logs.iter().rev().take(n) {
+        println!("{}", log)
+    }
+}
```
kate-goldenring, line 120: I believe that this needs to be changes to take the last n and keep the direction of ordering. I was thinking through this in this playground https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=efb52801edab3b4baa36ea2e68499d96
```suggestion
fn print_lastn_logs(logs: Vec<String>, n: usize) {
    for log in &logs[logs.len()-n..] {
        println!("{}", log)
    }
}
```

src/commands/mod.rs
kate-goldenring: yeah this sounds ideal. maybe we put up an issue for this and clean this up in a subsequent pr?

src/commands/logs.rs
```diff
@@ -0,0 +1,120 @@
...
+    #[clap(name = "follow", long = "follow")]
+    pub follow: bool,
+
+    /// Number of lines to show from the end of the logs
+    #[clap(name = "tail", long = "tail", default_value = "10")]
+    pub lastn: usize,
+
+    /// interval in secs to refresh logs from cloud
+    #[clap(name = "interval", long = "interval", default_value = "5")]
+    pub interval_secs: u64,
```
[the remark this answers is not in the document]
rajatjindal: i was hoping there is something in built by clap for this? for now I've added a validate_args function

src/commands/logs.rs
rajatjindal: done

src/commands/logs.rs
rajatjindal: hmm, I had to clone still. without that i am getting a `move` error

src/commands/logs.rs
rajatjindal: hm, why do you think we need this change?

src/commands/logs.rs
rajatjindal: 👍 done

src/commands/mod.rs
rajatjindal: 👍 https://github.com/fermyon/cloud-plugin/issues/73

src/commands/logs.rs
rajatjindal: sounds good. could you please help me come up with a msg here:

`error while finding internal handle to logs`
`internal error, current channel for the app not found`
`internal identifier for cloud app not found`

rajatjindal:
> Thanks for adding this Rajat! To @rylev's point, instead of getting all logs, can you update `channel_logs` to take in `max` aka count. The [`api_channels_id_logs_get`](https://github.com/fermyon/cloud-plugin/blob/643ceaca9a79ec334a93639079236ffc44f018f3/crates/cloud/src/client.rs#L289) already has support for requesting n many logs and it is the [latest n returned by journal](https://github.com/fermyon/journal/blob/c9fabef3f25331510e8152094b13cfe0339b818a/journal-api/journal/v1/journal.proto#L35).

I want to do that too. but lets say we start the tail command and fetched 100 lines of logs. now the app added 5 new line of logs. when we ask for 100 lines of logs again, how do we know that the last 5 are the latest ones to display?

src/commands/logs.rs
rajatjindal: 👍

src/commands/logs.rs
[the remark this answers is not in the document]
rajatjindal: 👍

src/commands/logs.rs
rajatjindal: 👍  great catch. thank you.

src/commands/logs.rs
rajatjindal: 🎉  thank you.

rajatjindal:
> > Thanks for adding this Rajat! To @rylev's point, instead of getting all logs, can you update `channel_logs` to take in `max` aka count. The [`api_channels_id_logs_get`](https://github.com/fermyon/cloud-plugin/blob/643ceaca9a79ec334a93639079236ffc44f018f3/crates/cloud/src/client.rs#L289) already has support for requesting n many logs and it is the [latest n returned by journal](https://github.com/fermyon/journal/blob/c9fabef3f25331510e8152094b13cfe0339b818a/journal-api/journal/v1/journal.proto#L35).
> 
> I want to do that too. but lets say we start the tail command and fetched 100 lines of logs. now the app added 5 new line of logs. when we ask for 100 lines of logs again, how do we know that the last 5 are the latest ones to display?

I looked at the logging api code. if we can add ability to say "give me logs since <timestamp>", then we might be able to circumvent the problem. I'll give that a try soon.

cc @michelleN - I know you were looking at the logging api as well.

rajatjindal:
this will fix #107

src/commands/logs.rs
rajatjindal: this is outdated code.

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+    #[clap(
+        name = "environment-name",
+        long = "environment-name",
+        env = DEPLOYMENT_ENV_NAME_ENV
+    )]
+    pub deployment_env_id: Option<String>,
+
+    /// App name
+    #[clap(name = "app-name", long = "app-name")]
+    pub app_name: String,
```
karthik2804, line 31: To be consistent with other cases we probably do not want to have an argument (eg) 'apps info' and `apps delete` commands. This would change to 

```bash
spin cloud logs <APP_NAME>
```

```suggestion
    pub app_name: String,
```

itowlson: If this is the same as the app name in `spin cloud link` and `spin cloud variables` then it should be `#[clap(short = 'a', long = "app")]` (actually we are not 100% consistent about having a `short` form but the long form is always "app").

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+    pub async fn run(self) -> Result<()> {
+        let client = create_cloud_client(self.deployment_env_id.as_deref()).await?;
+        let app_name: String = self.app_name.clone();
+
+        self.logs(client, app_name.as_str())
+            .await
+            .map_err(|e| anyhow!("{:?}\n\nLearn more at {}", e, "DEVELOPER_CLOUD_FAQ"))
+    }
+
+    async fn logs(self, client: CloudClient, app_name: &str) -> Result<()> {
```
itowlson, line 60: Not quite sure why you pass `app_name` separately here when it's the same as `self.app_name`?

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+        let app_id = match client
+            .get_app_id(app_name)
+            .await
+            .map_err(|_e| anyhow!("app with name {:?} not found", app_name))?
```
itowlson, line 64: This seems to lump all errors into the "app not found" case, which would point diagnosis in the wrong direction when the app _does_ exist but there's some other problem.  Consider propagating the original error with context, or at least logging it.

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+            self.since,
+        )
+        .await?;
+
+        Ok(())
+    }
+}
+
+async fn fetch_logs_and_print_loop(
+    client: &CloudClient,
```
itowlson, line 89:
```suggestion
    client: &impl CloudClientInterface,
```

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+    channel_id: Uuid,
+    follow: bool,
+    interval: u64,
```
itowlson, line 92: This should be a Duration by this point.

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    loop {
+        tokio::time::sleep(tokio::time::Duration::from_secs(interval)).await;
+        new_since =
+            fetch_logs_and_print_once(client, channel_id, None, new_since.to_owned()).await?;
+    }
+}
+
+async fn fetch_logs_and_print_once(
+    client: &CloudClient,
```
itowlson, line 112:
```suggestion
    client: &impl CloudClientInterface,
```

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+    channel_id: Uuid,
+    max_lines: Option<i32>,
+    since: String,
+) -> Result<String> {
+    let entries = client
+        .channel_logs_raw(channel_id.to_string(), max_lines, Some(since.to_string()))
```
itowlson, line 118: I guess this is out of scope for this PR but it's weird that we get the channel ID as a Uuid and then have to convert it back to a String to use it.  It would be nice to treat it _always_ as a Uuid.

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+
+async fn fetch_logs_and_print_loop(
+    client: &CloudClient,
+    channel_id: Uuid,
+    follow: bool,
+    interval: u64,
+    max_lines: Option<i32>,
+    since: Duration,
+) -> Result<()> {
+    let mut new_since = Utc::now().sub(since).to_rfc3339();
```
itowlson, line 96: `new` here is a bit of a wart - when you're going round the loop the it's not necessarily "new" any more.  I don't love shadowing but maybe plain "since" would be clearer?

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+        fetch_logs_and_print_once(client, channel_id, max_lines, new_since.to_owned()).await?;
+
+    if !follow {
+        return Ok(());
+    }
+
+    loop {
+        tokio::time::sleep(tokio::time::Duration::from_secs(interval)).await;
+        new_since =
+            fetch_logs_and_print_once(client, channel_id, None, new_since.to_owned()).await?;
```
itowlson, line 107: Is it necessary to keep calling `to_owned()` on what is already a `String`?

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+        return Ok(since.to_owned());
+    }
+
+    Ok(print_lastn_logs(&entries).to_owned())
+}
+
+fn print_lastn_logs(entries: &[Entry]) -> &str {
+    let mut new_since: &str = "";
+    for entry in entries.iter().rev() {
+        for line in entry.log_lines.as_ref().unwrap() {
```
itowlson, line 132: Comment why these unwraps are safe

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+        let value: u64 = parg.parse()?;
+        std::time::Duration::from_secs(value * 60)
+    } else if let Some(parg) = arg.strip_suffix('h') {
+        let value: u64 = parg.parse()?;
+        std::time::Duration::from_secs(value * 60 * 60)
+    } else if let Some(parg) = arg.strip_suffix('d') {
+        let value: u64 = parg.parse()?;
+        std::time::Duration::from_secs(value * 24 * 60 * 60)
+    } else {
+        return Err(anyhow!("invalid duration"));
```
itowlson, line 155: This will be displayed to a user who saw nothing wrong with "3min" or "1h30m".  Help them understand what the command accepts as "valid".

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    /// Follow logs output
+    #[clap(name = "follow", long = "follow")]
+    pub follow: bool,
+
+    /// Number of lines to show from the end of the logs
+    #[clap(name = "tail", long = "tail", default_value = "10")]
+    pub max_lines: i32,
+
+    /// interval in secs to refresh logs from cloud
```
itowlson, line 41: Consistent capitalisation
```suggestion
    /// Interval in secs to refresh logs from cloud
```

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+    #[clap(parse(try_from_str = parse_interval), name="interval", long="interval", default_value = "2")]
+    pub interval_secs: u64,
+
+    /// fetch logs since
```
itowlson, line 45: If I read this as a doc string I would think I should enter a date or time.  "Fetch logs since 2:30."  Provide guidance e.g.
```suggestion
    /// How long ago to fetch logs from, expressed as a duration. The duration format
    /// is a number and a unit, where the unit is 's' for seconds, 'm' for minutes, 'h' for hours
    /// or 'd' for days (e.g. "30m" for 30 minutes ago).  The default it 7 days.
```

(The start of that suggestion feels a bit too wordy and indirect, but hopefully gives an idea of the kind of thing.)

src/commands/logs.rs
```diff
@@ -0,0 +1,168 @@
...
+        for line in entry.log_lines.as_ref().unwrap() {
+            println!("{}", line.line.as_ref().unwrap());
+            new_since = line.time.as_ref().unwrap().as_str()
+        }
+    }
+
+    new_since
+}
+
+fn parse_duration(arg: &str) -> Result<std::time::Duration, anyhow::Error> {
```
itowlson, line 141: tiny nit: I find this form more concise:

```suggestion
fn parse_duration(arg: &str) -> anyhow::Result<std::time::Duration> {
```

src/commands/logs.rs
rajatjindal: in one of the iteration i was running into move error IIRC. changed it to use app_name from self.

src/commands/logs.rs
rajatjindal: done.

src/commands/logs.rs
rajatjindal:
```
cloud-plugin logs --app-name non-exist --since 30min
error: Invalid value "30min" for '--since <since>': supported formats are "300s", "5m", "4h" or "1d". formats such as "1h30m" or "30min" are not supported
```

is this looking ok now?

itowlson: I'd drop the "formats such as "1h30m" or "30min" are not supported" - you've told them what _is_ supported and that's probably all that's needed.  We can always expand the error if people do try combining units but simple will do to start with.  Thanks!

src/commands/logs.rs
rajatjindal:
```
Only return logs newer than a relative duration.
```

src/commands/logs.rs
rajatjindal: actually `interval` is different from `since`.  

interval is the time in seconds after which request is made to server to fetch latest logs. This is required as our logs api don't support streaming right now.

src/commands/logs.rs
rajatjindal: thanks. renamed variable to `app` and changed it to be an argument instead of flag

src/commands/logs.rs
itowlson: I know interval is different from since, but it's still a Duration.  In the same way as you get `since` out of a string and into a Duration before getting into the core operation, it would be good to get `interval` out of a `u64` and into a Duration too.  That avoids any uncertainty about the semantics of the u64.

src/commands/logs.rs
rajatjindal: done.

src/commands/logs.rs
rajatjindal: 👍  done. thanks Ivan. (just wanted to make sure the distinction between interval and since was clear in this context).

src/commands/logs.rs
rajatjindal: yeah, all functions in client currently use string. I guess the generated client will need some tweaking for this.

src/commands/logs.rs
rajatjindal: just to be clear, when we are going in the loop, new_since is updated after each loop iteration (to the last log timestamp), so it is indeed new_since. I can change it to since if you still think that would be better?

src/commands/logs.rs
rajatjindal: I think one of the previous comment had me remove `app_name` passed along as parameter. Before that change, we were getting a moved error with this. which is now no longer the case 🎉  thank you Ivan.

src/commands/logs.rs
itowlson: I can't muster strong feelings about it  _grin_ - feel free to leave it as is.

src/commands/logs.rs
rajatjindal: fixed now

src/commands/logs.rs
```diff
@@ -0,0 +1,175 @@
...
+        std::time::Duration::from_secs(value * 60)
+    } else if let Some(parg) = arg.strip_suffix('h') {
+        let value: u64 = parg.parse()?;
+        std::time::Duration::from_secs(value * 60 * 60)
+    } else if let Some(parg) = arg.strip_suffix('d') {
+        let value: u64 = parg.parse()?;
+        std::time::Duration::from_secs(value * 24 * 60 * 60)
+    } else {
+        return Err(anyhow!(
+            r#"supported formats are "300s", "5m", "4h" or "1d""#
```
kate-goldenring, line 161:
```suggestion
            r#"time format must be a number followed by an allowed unit ("300s", "5m", "4h" or "1d")"#
```

src/commands/logs.rs
```diff
@@ -0,0 +1,175 @@
...
+        self.logs(&client)
+            .await
+            .map_err(|e| anyhow!("{:?}\n\nLearn more at {}", e, DEVELOPER_CLOUD_FAQ))
+    }
+
+    async fn logs(self, client: &impl CloudClientInterface) -> Result<()> {
+        let app_id = match client
+            .get_app_id(&self.app)
+            .await
+            .map_err(|e| anyhow!("app with name {:?} not found\n\n{:?}", &self.app, e))?
```
kate-goldenring, line 63: Why the new lines here on the error? There aren't new lines on the next identical error. Can we be consistent?

src/commands/logs.rs
kate-goldenring: The `new` also caught me off guard. I think `curr_since` would be better.

src/commands/logs.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    Ok(print_lastn_logs(&entries).to_owned())
+}
+
+fn print_lastn_logs(entries: &[Entry]) -> &str {
+    let mut new_since: &str = "";
+    for entry in entries.iter().rev() {
+        for line in entry.log_lines.as_ref().unwrap() {
+            // line will always have some content, so it is safe to unwrap
+            println!("{}", line.line.as_ref().unwrap());
```
kate-goldenring, line 138: Can we prefix the log time here? It would be helpful to know the time of each log and make it easier to tell it is a log. For example, my logs for an app are the following which made me think the log command was erroring at first
```sh
Kates-MBP :: ~/Programs/cloud-plugin 2 » cargo run -- logs --since 10d boo-todo
    Finished dev [unoptimized + debuginfo] target(s) in 0.46s
     Running `target/debug/cloud-plugin logs --since 10d boo-todo`
Cannot find file /favicon.ico. Attempting fallback.
Cannot read file: cannot open /favicon.ico

Caused by:
    No such file or directory (os error 44)
```

kate-goldenring (commented):
Thanks for updating the APIs to make this possible @rajatjindal

src/commands/logs.rs
```diff
@@ -0,0 +1,175 @@
...
+use crate::commands::deploy::{DEVELOPER_CLOUD_FAQ, SPIN_DEPLOY_CHANNEL_NAME};
+use crate::opts::*;
+use clap::Parser;
+use uuid::Uuid;
+
+/// Fetch and tail logs of an application deployed to the Fermyon Cloud.
+#[derive(Parser, Debug)]
+#[clap(about = "fetch logs for an app from Fermyon Cloud")]
+pub struct LogsCommand {
+    /// Find app to fetch logs for in the Fermyon instance saved under the specified name.
```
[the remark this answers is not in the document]
rajatjindal: this is indeed for `deployment_env_id`. copied (and modified) from `deploy.rs` file

src/commands/logs.rs
rajatjindal: thank you

src/commands/logs.rs
rajatjindal: usually I try to follow pattern from existing code :) : https://github.com/fermyon/cloud-plugin/blob/d410545f4a62c81b911c33e49658cec30873be29/src/commands/deploy.rs#L125 

in this case, I copied this from `deploy.rs`. I am happy to change it to the recommended pattern.

src/commands/logs.rs
rajatjindal: I think we should add a flag to print timestamps with log line. e.g. there are instances where people have timestamps in their logs already, and duplicate timestamps won't look nice either.

if we agree to a flag, next question would be the default value. I don't have strong feelings about that. but I think we can show timestamps by default?

src/commands/logs.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    async fn logs(self, client: &impl CloudClientInterface) -> Result<()> {
+        let app_id = match client
+            .get_app_id(&self.app)
+            .await
+            .map_err(|e| anyhow!("app with name {:?} not found\n\n{:?}", &self.app, e))?
+        {
+            Some(x) => x,
+            None => return Err(anyhow!("app with name {:?} not found", &self.app)),
+        };
```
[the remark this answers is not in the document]
rajatjindal: getting error with change:

```
   Compiling cloud-plugin v0.4.0 (/Users/rajatjindal/go/src/github.com/fermyon/cloud-plugin)
error: expected one of `.`, `?`, `{`, or an operator, found `;`
  --> src/commands/logs.rs:64:87
   |
60 |         let app_id = match client
   |                      -----
   |                      |
   |                      while parsing this `match` expression
   |                      help: try removing this `match`
...
64 |             .map_err(|e| anyhow!("error finding app with name {:?}: {e}", &self.app))?;
   |                                                                                       ^ expected one of `.`, `?`, `{`, or an operator


```

tried removing `match`

```
   Compiling cloud-plugin v0.4.0 (/Users/rajatjindal/go/src/github.com/fermyon/cloud-plugin)
error[E0599]: no method named `map_err` found for enum `std::option::Option` in the current scope
  --> src/commands/logs.rs:64:14
   |
60 |           let app_id = client
   |  ______________________-
61 | |             .get_app_id(&self.app)
62 | |             .await
63 | |             .with_context(|| format!("app with name {:?} not found", &self.app))?
64 | |             .map_err(|e| anyhow!("error finding app with name {:?}: {e}", &self.app))?;
   | |             -^^^^^^^ help: there is a method with a similar name: `map_or`
```

src/commands/logs.rs
```diff
@@ -0,0 +1,175 @@
...
+        ));
+    };
+
+    Ok(duration)
+}
+
+fn parse_interval(arg: &str) -> anyhow::Result<std::time::Duration> {
+    let value = arg.parse()?;
+    if value < 2 {
+        return Err(anyhow!("interval cannot be less than 2 seconds"));
```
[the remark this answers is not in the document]
rajatjindal: done

src/commands/logs.rs
rajatjindal: 👍 , changing it to 'curr_since'

src/commands/logs.rs
rajatjindal: i completely did. added a if condition to check if this is defined. and changed the log_line to default to "" if its None

src/commands/logs.rs
itowlson: @rajatjindal that's specifically for the program-exit appending the cloud FAQ link on all deploy failures.  I don't think we generally split out underlying errors onto separate lines explicitly (though `context/with_context` render the "stack" of causes one per line).

src/commands/logs.rs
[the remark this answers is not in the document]
itowlson: The ambiguity arises in this case between "app saved under the specified name" vs "Fermyon instance saved under the specified name" (which doesn't arise in deploy, though I'm sure we could clarify the wording there too!).  How about:

```suggestion
    /// Use the Fermyon instance saved under the specified name.
```

src/commands/logs.rs
itowlson: The Cloud UI defaults to showing timestamps.  We may be better off following the download format than the UI though, as the UI uses a non-sortable timestamp format, and uses client local time.  (The download format uses sortable UTC time; the disadvantage is that the timestamps get long.)

src/commands/logs.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    Ok(print_lastn_logs(&entries).to_owned())
+}
+
+fn print_lastn_logs(entries: &[Entry]) -> &str {
+    let mut new_since: &str = "";
+    for entry in entries.iter().rev() {
+        for line in entry.log_lines.as_ref().unwrap() {
+            // line will always have some content, so it is safe to unwrap
```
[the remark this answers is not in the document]
rajatjindal: I have restructured this code a bit. happy to make further changes if there is a different rust idomatic way of doing this.

src/commands/logs.rs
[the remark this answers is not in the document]
rylev: Sorry - this is why I shouldn't try to write Rust without a compiler. Removing the `match` is right. It seems you switched the `with_context` and `map_err` lines. If you switch those back, does it work?

src/commands/logs.rs
rajatjindal: i have restructured this function a bit to check for None etc.

src/commands/logs.rs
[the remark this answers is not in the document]
rajatjindal: Ryan helped me fix it. done

src/main.rs
```diff
@@ -35,6 +36,8 @@ enum CloudCli {
     Deploy(DeployCommand),
     /// Login to Fermyon Cloud
     Login(LoginCommand),
+    /// Fetch app logs from Fermyon Cloud
```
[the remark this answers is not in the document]
rajatjindal: done. thanks Kate

rajatjindal:
@kate-goldenring @rylev @itowlson @karthik2804 

thank you for your feedback on this PR. I think this is now ready to be merged. please let me know if there are comments that I might have missed to address. 

thanks

src/commands/logs.rs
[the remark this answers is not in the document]
itowlson: @rajatjindal this seems to be marked as resolved without any change being made - is there a push pending?

src/commands/logs.rs
```diff
@@ -0,0 +1,203 @@
...
+                Some(time) => {
+                    if show_timestamp {
+                        println!("[{time}] {log}");
+                    } else {
+                        println!("{log}");
+                    }
+
+                    since = Some(time.as_str());
+                }
+                None => panic!("timestamp is None for the log entry"),
```
itowlson, line 168: This is not panic-worthy.

itowlson (commented):
I don't think anything I have is blocking (though it would be _highly_ desirable to get rid of that panic), but I don't feel I know enough to approve, sorry.

src/commands/logs.rs
```diff
@@ -0,0 +1,203 @@
...
+        default_value = "true",
+        action = clap::ArgAction::Set
+    )]
+    pub show_timestamp: bool,
+}
+
+impl LogsCommand {
+    pub async fn run(self) -> Result<()> {
+        let client = create_cloud_client(self.deployment_env_id.as_deref()).await?;
+        self.logs(&client).await.map_err(|e| anyhow!("{:?}", e))
```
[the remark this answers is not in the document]
rajatjindal: 👍 make sense. removed

src/commands/logs.rs
```diff
@@ -0,0 +1,203 @@
...
+        Ok(())
+    }
+}
+
+async fn fetch_logs_and_print_loop(
+    client: &impl CloudClientInterface,
+    channel_id: Uuid,
+    follow: bool,
+    interval: Duration,
+    max_lines: Option<i32>,
```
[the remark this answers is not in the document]
rajatjindal: down the line we need to pass None (when following logs). I have changed it to be `i32` until the function call where that is required

src/commands/logs.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    let updated_since = print_lastn_logs(&entries, show_timestamp);
+    if updated_since.is_some() {
+        return Ok(updated_since.unwrap().to_owned());
+    }
+
+    Ok(since)
+}
+
+fn print_lastn_logs(entries: &[Entry], show_timestamp: bool) -> Option<&str> {
```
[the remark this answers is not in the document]
rajatjindal: done

src/commands/logs.rs
rajatjindal: changed it to do nothing.

rajatjindal:
@itowlson, as always, thank you for your insightful feedback on this PR.

src/commands/logs.rs
```diff
@@ -0,0 +1,208 @@
...
+use crate::commands::create_cloud_client;
+use crate::commands::deploy::SPIN_DEPLOY_CHANNEL_NAME;
+use crate::opts::*;
+use clap::Parser;
+use uuid::Uuid;
+
+/// fetch logs for an app from Fermyon Cloud
+#[derive(Parser, Debug)]
+pub struct LogsCommand {
+    /// Find app to fetch logs for in the Fermyon instance saved under the specified name.
```
itowlson, line 19: @rajatjindal forgive me for nagging on this but I would strongly encourage changing this to something generic like 

```suggestion
    /// Use the Fermyon instance saved under the specified name.
```

The mention of "app" makes it sound like the app, a la "find the app blah blah the specified name"

src/commands/logs.rs
itowlson: Should it do nothing or record the log text without the timestamp?  (I don't know the correct behaviour: if "do nothing" is right then all good!)

itowlson:
Looking good @rajatjindal, just one question and one thing that I mentioned in a comment but forgot to log in the review - sorry about that!

src/commands/logs.rs
rajatjindal: thank you for being persistent. I had intended to address this feedback, but it seems like it got lost in other comments while i was going through the comments.

src/commands/logs.rs
rajatjindal: if we do print log text here without updating `since`, then we may endup in a situation where this particular line can keep printing again and again (until we get a valid entry which endup updating `since`). 

my intution is that if api returned a valid entry, it will always have both text and time, so in this case, we can ignore the log text.

bacongobbler:
Would you mind changing this to use the /api/apps/:id/logs endpoints? The latest version of the cloud-openapi client contains these endpoints. The channel log endpoints are planned to be deprecated soon.

rajatjindal:
I have update openapi client and updated the PR to use new endpoints.

radu-matei (approved):
Tested this end to end and everything looks great!

LGTM

bacongobbler (approved):

src/commands/logs.rs
```diff
@@ -0,0 +1,192 @@
...
+                    if show_timestamp {
+                        println!("[{time}] {log}");
+                    } else {
+                        println!("{log}");
+                    }
+
+                    since = Some(time.as_str());
+                }
+                None => (),
+            }
```
rylev, line 158:
```suggestion
            if let Some(time) =  &log_entry.time {
                    if show_timestamp {
                        println!("[{time}] {log}");
                    } else {
                        println!("{log}");
                    }

                    since = Some(time.as_str());
            }
```

src/commands/logs.rs
```diff
@@ -0,0 +1,192 @@
...
+    let mut since = None;
+    for entry in entries.iter().rev() {
+        let Some(log_lines) = entry.log_lines.as_ref() else {
+            continue;
+        };
+
+        for log_entry in log_lines {
+            let Some(log) = log_entry.line.as_ref() else {
+                continue;
+            };
```
rylev, line 145:
```suggestion
        for log_entry in log_lines.iter().filter_map(|l| l.line.as_ref()) {
```

src/commands/logs.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    Ok(since)
+}
+
+fn print_logs(entries: &[Entry], show_timestamp: bool) -> Option<&str> {
+    let mut since = None;
+    for entry in entries.iter().rev() {
+        let Some(log_lines) = entry.log_lines.as_ref() else {
+            continue;
+        };
```
rylev, line 140:
```suggestion
    for entry in entries.iter().rev().filter_map(|e| e.long_lines.as_ref()) {
```

rylev (approved):
Just some last second code nits. Feel free to merge this without addressing those!

kate-goldenring (approved):
LGTM! Thanks @rajatjindal

src/commands/logs.rs
rajatjindal: i tried this and ran into a problem (needed to use filter instead of filter_map). tried spending sometime to fix that, but the code seemed more readable in current form. so going to skip this one for now

src/commands/logs.rs
rajatjindal: i tried this and ran into a problem (needed to use filter instead of filter_map). tried spending sometime to fix that, but the code seemed more readable in current form. so going to skip this one for now

PR #60, opened 2023-08-17
ci(*): add makefile

vdice:
This is a little bit of a smorgasbord of CI additions and am open to scoping to just the ones desired.  My primary motivation was to reduce the build->package->install cycle during development, hence the Makefile and associated targets, doing away with the need to manually edit cloud.json and install, etc.  I figured while doing so, I would add lint and test targets and use these in CI so that devs and automation are invoking the same commands.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,13 +44,8 @@ jobs:
 
       - uses: actions/checkout@v3
 
-      - name: Cargo Format
-        run:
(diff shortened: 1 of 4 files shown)


vdice:
With https://github.com/fermyon/cloud-plugin/pull/72 the local dev loop is much improved.  I'll close this for now and perhaps we can return with some CI improvements/consolidation in the near future.

PR #67, opened 2023-08-18
add apps subcommand that allows to list and delete apps

### LINKED ISSUE #9: Support listing all application in the cloud
In reference to #8. A command to list all application in the cloud, would be good, to easily retrieve a name of an existing application. E.g., `spin cloud app list`.

### LINKED ISSUE #40: Spin CLI should allow me to remove a deployment
**Is your feature request related to a problem? Please describe.**
I can't clean up a deployment I created within the workflow for the CLI

**Describe the solution you'd like**
I would like to be able to run "spin destroy" and have my deployment removed from Fermyon Cloud

**Describe alternatives you've considered**
The only alternative is logging into the UI and doing it manually today. A plugin could be created to handle this (and other cloud tasks like listing, etc). I think this should either be in core CLI or all the cloud stuff including this moved off.

karthik2804:
This PR adds a new top-level command `apps` that further has `list` and `delete` subcommands. 

fixes #9
fixes #40

### DIFF
--- a/src/commands/apps.rs
+++ b/src/commands/apps.rs
@@ -1,0 +1,80 @@
+use crate::commands::{client_and_app_id, create_cloud_client};
+use crate::opts::*;
+use anyhow::{Context, Result};
+use clap::{Args, Parser};
+use cloud_openapi::models::AppItemPage;
+
+#[derive(Parser, Debug)]
+#[clap(about = "Manage applications deployed to Fermyon Cloud")]
+pub enum AppsCommand {
+    /// List all the apps deployed in Fermyon Cloud
+    List(ListCommand),
+    /// Delete an app deployed in Fermyon Cloud
+    Delete(DeleteCommand),
+}
+
+#[derive(Parser, Debug)]
+pub struct ListCommand {
+    #[clap(flatten)]
+    common: CommonArgs,
+}
+
+#[derive(Parser, Debug)]
+pub struct DeleteCommand {
+    /// Name of Spin app
+    pub app: String,
+    #[clap(flatten)]
+    common: CommonArgs,
+}
+
+#[derive(Debug, Default, Args)]
+struct CommonArgs {
+    /// Deploy to the Fermyon instance saved under the specified name.
+    /// If omitted, Spin deploys to the default unnamed instance.
+    #[clap(
+        name = "environment-name",
+        long = "environment-name",
+        env = DEPLOYMENT_ENV_NAME_ENV
+    )]
+    pub deployment_env_id: Option<String>,
+}
+
+impl AppsCommand {
+    pub async fn run(self) -> Result<()> {
+        match self {
+            AppsCommand::List(cmd) => {
+                let client = create_cloud_client(cmd.common.deployment_env_id.as_deref()).await?;
+                let mut app_list_page = client.list_apps(DEFAULT_APPLIST_PAGE_SIZE, None).await?;
+                if app_list_page.total_items > 0 {
+                    eprintln!("No applications found");
+                } else {
+                    print_app_list(&app_list_page);
+                    let mut page_index = 1;
+                    while !app_list_page.is_last_page {
+                        app_list_page = client
+                            .list_apps(DEFAULT_APPLIST_PAGE_SIZE, Some(page_index))
+                            .await?;
+                        print_app_list(&app_list_page);
+                        page_index += 1;
+                    }
+                }
+            }
+            AppsCommand::Delete(cmd) => {
+                let (client, app_id) =
+                    client_and_app_id(cmd.common.deployment_env_id.as_deref(), &cmd.app).await?;
+                client
+                    .remove_app(app_id.to_string())
+                    .await
+                    .with_context(|| format!("Problem deleting app named {}", &cmd.app))?;
+                println!("Deleted app \"{}\" successfully.", &cmd.app);
+            }
+        }
+        Ok(())
+    }
+}
+
+fn print_app_list(page: &AppItemPage) {
+    for app in &page.items {
+        println!("{}", app.name);
+    }
+}
--- a/src/commands/mod.rs
+++ b/src/commands/mod.rs
@@ -1,10 +1,11 @@
+pub mod apps;
 pub mod deploy;
 pub mod login;
 pub mod sqlite;
 pub mod variables;
 
-use crate::commands::deploy::login_connection;
-use anyhow::{bail, Result};
+use crate::{commands::deploy::login_connection, opts::DEFAULT_APPLIST_PAGE_SIZE};
+use anyhow::{bail, Context, Result};
 use cloud::client::{Client as CloudClient, ConnectionConfig};
 use uuid::Uuid;
 
@@ -19,10 +20,21 @@ pub(crate) async fn create_cloud_client(deployment_env_id: Option<&str>) -> Resu
 }
 
 pub(crate) async fn get_app_id_cloud(cloud_client: &CloudClient, name: &str) -> Result<Uuid> {
-    let apps_vm = CloudClient::list_apps(cloud_client).await?;
+    let apps_vm = CloudClient::list_apps(cloud_client, DEFAULT_APPLIST_PAGE_SIZE, None).await?;
     let app = apps_vm.items.iter().find(|&x| x.name == name);
     match app {
         Some(a) => Ok(a.id),
         None => bail!("No app with name: {}", name),
     }
+}
+
+async fn client_and_app_id(
+    deployment_env_id: Option<&str>,
+    app: &str,
+) -> Result<(CloudClient, Uuid)> {
+    let client = create_cloud_client(deployment_env_id).await?;
+    let app_id = get_app_id_cloud(&client, app)
+        .await
+        .with_context(|| format!("Could not find app_id for app {}", app))?;
+    Ok((client, app_id))
 }
--- a/crates/cloud/src/client.rs
+++ b/crates/cloud/src/client.rs
@@ -155,10 +155,18 @@ impl Client {
             .map_err(format_response_error)
     }
 
-    pub async fn list_apps(&self) -> Result<AppItemPage> {
-        api_apps_get(&self.configuration, None, None, None, None, None, None)
-            .await
-            .map_err(format_response_error)
+    pub async fn list_apps(&self, page_size: i32, page_index: Option<i32>) -> Result<AppItemPage> {
+        api_apps_get(
+            &self.configuration,
+            None,
+            page_index,
+            Some(page_size),
+            None,
+            None,
+            None,
+        )
+        .await
+        .map_err(format_response_error)
     }
 
     pub async fn get_channel_by_id(&self, id: &str) -> Result<ChannelItem> {
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -404,7 +404,7 @@ impl DeployCommand {
         cloud_client: &CloudClient,
         name: String,
     ) -> Result<Option<Uuid>> {
-        let apps_vm = CloudClient::list_apps(cloud_client).await?;
+        let apps_vm = CloudClient::list_apps(cloud_client, DEFAULT_APPLIST_PAGE_SIZE, None).await?;
         let app = apps_vm.items.iter().find(|&x| x.name == name.clone());
         match app {
             Some(a) => Ok(Some(a.id)),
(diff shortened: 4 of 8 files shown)


src/commands/apps.rs
```diff
@@ -0,0 +1,67 @@
...
+    )]
+    pub deployment_env_id: Option<String>,
+}
+
+impl AppsCommand {
+    pub async fn run(self) -> Result<()> {
+        match self {
+            AppsCommand::List(cmd) => {
+                let client = create_cloud_client(cmd.common.deployment_env_id.as_deref()).await?;
+                let app_list_page = client.list_apps().await?;
```
karthik2804, line 46: This just fetches the first 50 apps by default. I am not sure what the UX for this should be. Do we want to follow the behavior of something like `ls` and print the entire list?

kate-goldenring: For others to reference, default is set in [Nimbus here](https://github.com/fermyon/nimbus/blob/main/src/Api/Controllers/AppsController.cs#L44-L45). 
Maybe provide an `--all` flag that will cause all to be printed. This would require making a `list_all_apps` fn that will make `total_items` / 50 calls.

src/commands/apps.rs
```diff
@@ -0,0 +1,67 @@
...
+                let app_list: Vec<&str> = app_list_page
```
kate-goldenring, line 47: You can consume the `app_list_page` object here. (sorry suggestion wasn't working)
```sh
 let app_list: Vec<String> = app_list_page
                    .items
                    .into_iter()
                    .map(|app_item| app_item.name)
                    .collect();
```

src/commands/mod.rs
```diff
@@ -26,3 +27,14 @@ pub(crate) async fn get_app_id_cloud(cloud_client: &CloudClient, name: &str) ->
...
     }
 }
+
+async fn client_and_app_id(
+    deployment_env_id: Option<&str>,
+    app: &str,
+) -> Result<(CloudClient, Uuid)> {
+    let client = create_cloud_client(deployment_env_id).await?;
+    let app_id = get_app_id_cloud(&client, app)
+        .await
```
kate-goldenring, line 37: I think most of the functions in `mod` we are hoping to move to the `cloud` crate. I don't think this one is powerful enough to move there and for that reason i would say we should keep this in `variables.rs` and just use the two lines in `apps.rs`

src/commands/apps.rs
kate-goldenring: But I think something like this would be nice, which has a header and a total count and then just line by line like you have
```sh
spin cloud apps list
Apps (3)
foo
bar
fee
```

src/commands/mod.rs
karthik2804: I think the function will be potentially used multiple times within `apps.rs` in case we add commands to get information about one particular app. something along the lines of 

```
spin cloud apps details my-app
```

src/commands/apps.rs
karthik2804: Would be a bad idea to change the `page_size` to the size of `total_items` which would help us avoid making multiple calls?

karthik2804: On second thought - it would not make much of a big difference as the current maximum limit is 100 and that would be 2 requests regardless if there are over 50 apps.

itowlson: It would be surprising if `--all` wasn't the default.

We already did this for channels and revisions (https://github.com/fermyon/spin/pull/973) and it was pretty simple - presumably apps would be very much the same?

@kate-goldenring The initial line could be a pain for piping e.g. `spin cloud apps list | sort`.

src/commands/apps.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+impl AppsCommand {
+    pub async fn run(self) -> Result<()> {
+        match self {
+            AppsCommand::List(cmd) => {
+                let client = create_cloud_client(cmd.common.deployment_env_id.as_deref()).await?;
+                let mut app_list_page = client.list_apps(DEFAULT_APPLIST_PAGE_SIZE, None).await?;
+                match (app_list_page.is_last_page, cmd.all) {
+                    (true, _) | (false, true) => println!("Apps ({})", app_list_page.total_items),
```
itowlson, line 52: This plays merry hob with piping e.g. `spin cloud apps list | sort`

src/commands/apps.rs
```diff
@@ -0,0 +1,89 @@
...
+                    (false, false) => println!("Note: Displaying the first {} of {} Apps. To list all the apps run \"spin cloud apps list -all\"", DEFAULT_APPLIST_PAGE_SIZE, app_list_page.total_items),
```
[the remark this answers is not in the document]
karthik2804: I like making it the default option - I will add an option called count that returns the total number of apps.

src/commands/apps.rs
kate-goldenring: True unless you do `spin cloud apps list | tail -n +2 | sort`. We might want to standardize what to do about listed output. Should we never add a header row? When is it useful? With `sqlite` CLI sketched, we've talked about having a header.

src/commands/apps.rs
karthik2804: Removed it.

src/commands/apps.rs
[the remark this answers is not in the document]
karthik2804: I have removed the `--all` flag - therefore this has been removed.

src/commands/apps.rs
itowlson: For an unhelpful comparison:

`spin plugins list` prints a bare list, no header row.

`spin templates list` prints a table with headers and separators and borders and everything.  The reason for this is to display multiple fields (name, description, installed from).

I'm not hugely sold on the usefulness of a header row that just shows the object type and a count.  I think the argument for the `sqlite` case was so the output wasn't empty if they had no databases.  But in that case it might be clearer just to print "You have no applications / databases / etc."?  (maybe to stderr instead of stdout, so that pipes could tell the cases apart)

src/commands/apps.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Delete an app deployed in Fermyon Cloud
+    Delete(DeleteCommand),
+}
+
+#[derive(Parser, Debug)]
+pub struct ListCommand {
+    #[clap(flatten)]
+    common: CommonArgs,
+    /// Lists all the apps.
+    #[clap(short = 'a', long = "all", takes_value = false)]
```
itowlson, line 21: This syntax was for an old option

src/commands/apps.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub count: bool,
```
itowlson, line 22: I am not convinced this is particularly useful; I'd consider taking it out, and adding it back if someone says "I want to count my apps more efficiently than list and `wc`".

In that case I give you permission to call the PR "I told you so"

src/commands/apps.rs
```diff
@@ -0,0 +1,86 @@
...
+        match self {
+            AppsCommand::List(cmd) => {
+                let client = create_cloud_client(cmd.common.deployment_env_id.as_deref()).await?;
+                let mut app_list_page = client.list_apps(DEFAULT_APPLIST_PAGE_SIZE, None).await?;
+                match cmd.count {
+                    true => {
+                        println!("Count: {}", app_list_page.total_items);
+                    }
+                    false => {
+                        print_app_list(&app_list_page);
```
itowlson, line 56: Consider eprintln-ing something like "You have no applications" if the count is 0.

itowlson (changes requested):
The only blocker is the incorrect syntax for the count flag, but I think we should also make sure we have consensus on the UI e.g. whether we want to have the count flag, whether we want to have a header row, and whether we want to print anything when there are no entities.  This is important because of likely broader implications e.g. we'd probably want the `sqlite list` command to have the same behaviour.

src/commands/apps.rs
karthik2804: Removed and noted!

src/commands/apps.rs
karthik2804: Removed it.

src/commands/apps.rs
karthik2804: Done!

src/commands/apps.rs
kate-goldenring: That's a good point about a header being improperly used to denote empty list.

src/commands/apps.rs
```diff
@@ -0,0 +1,83 @@
...
+    pub deployment_env_id: Option<String>,
+}
+
+impl AppsCommand {
+    pub async fn run(self) -> Result<()> {
+        match self {
+            AppsCommand::List(cmd) => {
+                let client = create_cloud_client(cmd.common.deployment_env_id.as_deref()).await?;
+                let mut app_list_page = client.list_apps(DEFAULT_APPLIST_PAGE_SIZE, None).await?;
+                match app_list_page.total_items > 0 {
```
kate-goldenring, line 48: Curious, why did you choose to do a match statement here instead of an if statement?

src/commands/apps.rs
```diff
@@ -0,0 +1,83 @@
...
+                }
+            }
+            AppsCommand::Delete(cmd) => {
+                let (client, app_id) =
+                    client_and_app_id(cmd.common.deployment_env_id.as_deref(), &cmd.app).await?;
+                client
+                    .remove_app(app_id.to_string())
+                    .await
+                    .with_context(|| format!("Problem deleting app named {}", &cmd.app))?;
+                println!("Delete app named \"{}\" Successully!", &cmd.app);
```
kate-goldenring, line 72:
```suggestion
                println!("Deleted app \"{}\" successfully.", &cmd.app);
```

kate-goldenring (commented):
I think this is looking ready to switch from Draft to Ready for Review. To your comment @itowlson, I am aligned with no headers and instead output on empty as is done here.

src/commands/apps.rs
karthik2804: Past reviews have trained me to prefer `match` over `if` but I do not know if it is not appropriate here.

kate-goldenring: i would usually only match on a bool if it was evaluating to a result that the match was capturing. I'm curious if there is a best practice here. @itowlson any thoughts?

itowlson: For a predicate like this I'd probably choose `if` because it avoids the clutter of the `true => / false =>` prefixes (but my opinion is not strong, and I am a really poor bellwether for what the Rust Gods consider best practice).

src/commands/apps.rs
```diff
@@ -0,0 +1,83 @@
...
+        long = "environment-name",
+        env = DEPLOYMENT_ENV_NAME_ENV
+    )]
+    pub deployment_env_id: Option<String>,
+}
+
+impl AppsCommand {
+    pub async fn run(self) -> Result<()> {
+        match self {
+            AppsCommand::List(cmd) => {
```
itowlson, line 45: Style suggestion: In many places we move each leaf command's `run` implementation down to that leaf command, so the intermediate level becomes (cf. https://github.com/fermyon/spin/blob/60011faf310e673f20fc7d5de20f4df8a99374e7/src/commands/plugins.rs#L43)

```
impl FieCommand {
  fn run(&self) {
    match self {
      Self::AliceCommand(cmd) => cmd.run(),
      Self::BobCommand(cmd) => cmd.run(),
    }
  }
}
```

This removes a level of structure and indentation from the actual real command logic, and keeps the implementations of the subcommands fully distinct in the source code.

It's not a must-do but it's a pattern I find significantly clearer.

itowlson (approved):
LGTM, one non-blocking suggestion (plus the comment about if vs match which is also non-blocking)

src/commands/apps.rs
karthik2804: I based it off the structure in `variables` and `sqlite`. I do not have a preference here. I am happy to go eitherway. Let me know!

itowlson: Oh, if `cloud-plugin` is inlining everything then I guess be consistent.

karthik2804:
@kate-goldenring @itowlson any strong opinions on `match` vs `if`? I am happy to go either way. I can update and then merge.

itowlson:
@karthik2804 I am Team If in this case

karthik2804:
@itowlson Well I will take the side with Team if this time around then!

itowlson:
image

PR #68, opened 2023-08-18
update user agent to include spin version

### LINKED ISSUE #14: Improve user-agent
Currently, Spin appears to be - [from code](https://github.com/fermyon/spin/blob/bcf66a7d5f42774fbe7949a1f7417b3541d8b334/crates/cloud/src/client.rs#L55) - sending e.g. `cloud/1.3.0` as the user-agent. This plugin will also - [currently](https://github.com/fermyon/cloud-plugin/blob/7acb2194fb498426c6656ab067e99aec2e0c9fb8/crates/cloud/src/client.rs#L56-L60) - be sending `cloud/*`. It would be nice if it sent something more detailed, like `cloud-plugin/${CARGO_PKG_VERSION} spin/${SPIN_VERSION}` (`SPIN_VERSION` is set by the plugin runner).

karthik2804:
closes #14

### DIFF
--- a/crates/cloud/src/client.rs
+++ b/crates/cloud/src/client.rs
@@ -62,9 +62,10 @@ impl Client {
         let configuration = Configuration {
             base_path,
             user_agent: Some(format!(
-                "{}/{}",
+                "{}/{} spin/{}",
                 env!("CARGO_PKG_NAME"),
-                env!("CARGO_PKG_VERSION")
+                env!("CARGO_PKG_VERSION"),
+                std::env::var("SPIN_VERSION").unwrap_or_else(|_| "0".to_string())
             )),
             client: reqwest::Client::builder()
                 .danger_accept_invalid_certs(conn_info.insecure)


crates/cloud/src/client.rs
```diff
@@ -62,9 +62,10 @@ impl Client {
         let configuration = Configuration {
             base_path,
             user_agent: Some(format!(
-                "{}/{}",
+                "{}/{} spin/{}",
                 env!("CARGO_PKG_NAME"),
-                env!("CARGO_PKG_VERSION")
+                env!("CARGO_PKG_VERSION"),
+                std::env::var("SPIN_VERSION").unwrap_or_default()
```
lann, line 68: I would maybe do this to help out UA parsers that expect something after the `/`:

```suggestion
                std::env::var("SPIN_VERSION").unwrap_or_else(|| "0".to_string())
```

lann (approved):
Minor nitty suggestion

PR #71, opened 2023-08-24
fix listing apps conditional

adamreese:


src/commands/apps.rs
```diff
@@ -45,7 +45,7 @@ impl AppsCommand {
             AppsCommand::List(cmd) => {
                 let client = create_cloud_client(cmd.common.deployment_env_id.as_deref()).await?;
                 let mut app_list_page = client.list_apps(DEFAULT_APPLIST_PAGE_SIZE, None).await?;
-                if app_list_page.total_items > 0 {
+                if app_list_page.total_items < 0 {
```
karthik2804, line 48: Can be equal to zero as well 
```suggestion
                if app_list_page.total_items <= 0 {
```

adamreese: good catch

itowlson (approved):

karthik2804 (approved):

Issue #84, opened 2023-09-14
CI: delete/recreate canary only when all artifacts are ready


vdice:
Currently, the [release.yml](https://github.com/fermyon/cloud-plugin/blob/main/.github/workflows/release.yml) workflow deletes and recreates the `canary` release first thing, prior to any of the artifacts being ready for attaching.  This presents a lengthy window when there are no binaries or `cloud.json` files attached to the release.  This presents problems as the [install script](https://github.com/fermyon/developer/blob/e734719e2d5d755e1e58890d586526f65b844c21/downloads/install.sh) for Spin can be used to specify that `canary` is the desired Spin release, along with all of the default plugins, which includes this one.  If the script is run while CI is in progress, some of the expected assets may not be available and thus the script may error out.  (See also https://github.com/fermyon/spin/issues/1771)

Update the workflow such that the canary release is only recreated once all assets are ready to be uploaded.  There may still be a window of exposure here, but at least it will be much shorter.  An effort to fully eliminate, such as using a CDN, is outside of the scope of this ticket.

PR #85, opened 2023-09-14
ci(release.yml): only recreate canary once all artifacts are ready to upload

vdice:
Closes https://github.com/fermyon/cloud-plugin/issues/84

Run on fork: https://github.com/vdice/cloud-plugin/actions/runs/6190247031

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,43 +23,9 @@ env:
   RUST_VERSION: 1.68
 
 jobs:
-  reset_canary_release: 
-    name: Delete and create Canary Release
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-
-      - name: 'Check if canary tag exists'
-        id: canaryExists
-        shell: bash
-        run: |
-          git fetch --prune --unshallow --tags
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -215,25 +180,62 @@ jobs:
...
+          tag_name: canary
+          delete_release: true
+
+      - name: Recreate canary tag and release
+        uses: ncipollo/release-action@v1.12.0
+        with:
+          tag: canary
+          allowUpdates: true
+          prerelease: true
+          artifacts: "cloud*.tar.gz,cloud.json,checksums-canary.txt"
```
karthik2804, line 237: I think the downloaded artifacts are nested in a folder. so it would have to be `cloud.json/cloud.json` and such.

vdice: Yea, I thought so as well at first, but I emulated what [fermyon/spin does](https://github.com/fermyon/spin/blob/main/.github/workflows/release.yml); appears to work on my fork: https://github.com/vdice/cloud-plugin/actions/runs/6190247031

karthik2804 (approved):

PR #94, opened 2023-09-19
Add get app subcommand to get app details

karthik2804:
This PR adds a get subcommand to apps to allow for getting details about an app. It helps with getting the Base URL of apps quickly without having to open up the cloud UI.

### DIFF
--- a/src/commands/apps.rs
+++ b/src/commands/apps.rs
@@ -2,7 +2,7 @@ use crate::commands::{client_and_app_id, create_cloud_client};
 use crate::opts::*;
 use anyhow::{Context, Result};
 use clap::{Args, Parser};
-use cloud_openapi::models::AppItemPage;
+use cloud_openapi::models::{AppItemPage, DomainValidationStatus};
 
 #[derive(Parser, Debug)]
 #[clap(about = "Manage applications deployed to Fermyon Cloud")]
@@ -11,6 +11,8 @@ pub enum AppsCommand {
     List(ListCommand),
     /// Delete an app deployed in Fermyon Cloud
     Delete(DeleteCommand),
+    /// Get Details about a deployed app in Fermyon Cloud
+    Get(GetCommand),
 }
 
 #[derive(Parser, Debug)]
@@ -21,6 +23,14 @@ pub struct ListCommand {
 
 #[derive(Parser, Debug)]
 pub struct DeleteCommand {
+    /// Name of Spin app
+    pub app: String,
+    #[clap(flatten)]
+    common: CommonArgs,
+}
+
+#[derive(Parser, Debug)]
+pub struct GetCommand {
     /// Name of Spin app
     pub app: String,
     #[clap(flatten)]
@@ -67,6 +77,41 @@ impl AppsCommand {
                     .await
                     .with_context(|| format!("Problem deleting app named {}", &cmd.app))?;
                 println!("Deleted app \"{}\" successfully.", &cmd.app);
+            }
+            AppsCommand::Get(cmd) => {
+                let (client, app_id) =
+                    client_and_app_id(cmd.common.deployment_env_id.as_deref(), &cmd.app).await?;
+                let app = client
+                    .get_app(app_id.to_string())
+                    .await
+                    .with_context(|| format!("Error: could not get details about {}", &cmd.app))?;
+
+                println!("Name: {}", &app.name);
+                if let Some(description) = &app.description {
+                    if !description.is_empty() {
+                        println!("Description: {}", description);
+                    }
+                }
+                match app.domain {
+                    Some(val) => {
+                        match val.validation_status {
+                            DomainValidationStatus::InProgress => {
+                                if let Some(url) = app.channels[0].domain.as_ref() {
+                                    println!("URL: https://{}", url);
+                                    println!("Validation for {} is in progress", val.name)
+                                }
+                            }
+                            DomainValidationStatus::Ready => {
+                                println!("URL: https://{}", val.name);
+                            }
+                        };
+                    }
+                    None => {
+                        if let Some(url) = app.channels[0].domain.as_ref() {
+                            println!("URL: https://{}", url);
+                        }
+                    }
+                };
             }
         }
         Ok(())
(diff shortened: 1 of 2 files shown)


src/commands/apps.rs
```diff
@@ -68,6 +78,29 @@ impl AppsCommand {
...
+                let app = client
+                    .get_app(app_id.to_string())
+                    .await
+                    .with_context(|| format!("Error: could not get details about {}", &cmd.app))?;
+
+                println!("Name: {}", &app.name);
+                println!("Description: {}", &app.description.unwrap_or_default());
+                println!(
+                    "URL: https://{}",
+                    app.channels[0].domain.as_ref().unwrap_or(&String::new())
```
itowlson, line 93: If there's no URL then skip this print rather than printing `URL: https://` (I know, it's a "can't happen" situation)

src/commands/apps.rs
```diff
@@ -68,6 +78,29 @@ impl AppsCommand {
...
+                );
+                if let Some(val) = app.domain {
+                    let status = match val.validation_status {
+                        cloud_openapi::models::DomainValidationStatus::InProgress => "In progress",
+                        cloud_openapi::models::DomainValidationStatus::Ready => "Ready",
+                    };
+                    println!("Custom domain: {}", val.name);
```
itowlson, line 100: If the custom domain is ready, wouldn't the URL already be showing it?  Alternative suggestion is to have URL be where the app is _currently_ available (whether custom domain or Fermyon custom subdomain or default deployment domain) and print an additional message _only if_ there is a custom domain pending.  That is:

* If there is no custom domain: show URL as the Fermyon URL
* If there is a custom domain and it is ready: show URL as the custom URL
* If there is a custom domain, but it is pending: show URL as the Fermyon one (the one which works!), and print an additional message along the lines of "Custom domain ... in progress / waiting to be provisioned / whatever"

src/commands/apps.rs
```diff
@@ -68,6 +78,29 @@ impl AppsCommand {
...
+            AppsCommand::Get(cmd) => {
+                let (client, app_id) =
+                    client_and_app_id(cmd.common.deployment_env_id.as_deref(), &cmd.app).await?;
+                let app = client
+                    .get_app(app_id.to_string())
+                    .await
+                    .with_context(|| format!("Error: could not get details about {}", &cmd.app))?;
+
+                println!("Name: {}", &app.name);
+                println!("Description: {}", &app.description.unwrap_or_default());
```
itowlson, line 90: If the description is empty, then skip this line rather than printing `Description:` with nothing next to it.

itowlson (changes requested):
The blocking one is the blank Description thing; the others are merely suggestions

src/commands/apps.rs
```diff
@@ -68,6 +78,29 @@ impl AppsCommand {
...
+                println!(
+                    "URL: https://{}",
+                    app.channels[0].domain.as_ref().unwrap_or(&String::new())
+                );
+                if let Some(val) = app.domain {
+                    let status = match val.validation_status {
+                        cloud_openapi::models::DomainValidationStatus::InProgress => "In progress",
```
kate-goldenring, line 97: nit: I would import `use cloud_openapi::models:DomainValidationStatus` instead of using the fully qualified path here
```suggestion
                        DomainValidationStatus::InProgress => "In progress",
```

src/commands/apps.rs
```diff
@@ -11,6 +11,8 @@ pub enum AppsCommand {
     List(ListCommand),
     /// Delete an app deployed in Fermyon Cloud
     Delete(DeleteCommand),
+    /// Get Details about a deployed app in Fermyon Cloud
+    Get(GetCommand),
```
kate-goldenring, line 15: `Get` seems not quite the right name to me. What about `Describe` or `Info`? Maybe `Get` is right here. Is there a CLI you've seen that you're modeling after?

karthik2804: I was just using the HTTP method here and it looked like it was descriptive enough of "get" app info - I am open to suggestions here. 
One possibility might be `get-info`?

karthik2804:
@itowlson I have addressed the comments.

src/commands/apps.rs
karthik2804: Done

src/commands/apps.rs
```diff
@@ -68,6 +78,41 @@ impl AppsCommand {
...
+                    if !description.is_empty() {
+                        println!("Description: {}", description);
+                    }
+                }
+                match app.domain {
+                    Some(val) => {
+                        match val.validation_status {
+                            DomainValidationStatus::InProgress => {
+                                if let Some(url) = app.channels[0].domain.as_ref() {
+                                    println!("URL: https://{}", url);
```
itowlson, line 100: wow, I let _one_ of my error messages stray into column 81, and `cargo fmt` gets mediaeval on my code with a rusty fork, but you have 9 levels of indent and it's all "right this way Mister Karthik can I escort you to the VIP zone".  THE INJUSTICE

More seriously... I am appreciating the additional refinement of the UI, but the downside is it has made this function very long and very deeply indented.  I'd strongly encourage breaking it up - ideally into command handlers, but at least pulling out some of the analysis and print logic into helpers.  It's not a blocker though.

itowlson (approved):
LGTM subject to agreement with Kate on the command name - thanks!

src/commands/apps.rs
```diff
@@ -11,6 +11,8 @@ pub enum AppsCommand {
     List(ListCommand),
     /// Delete an app deployed in Fermyon Cloud
     Delete(DeleteCommand),
+    /// Get Details about a deployed app in Fermyon Cloud
```
kate-goldenring, line 14: nit
```suggestion
    /// Get details about a deployed app in Fermyon Cloud
```

src/commands/apps.rs
kate-goldenring: Why `get-info` over just `info`? Maybe we take a vote in chat? I feel like to me, the following is displaying info, and `spin cloud apps info "lottery"` feels clearer. "get" like you mention is informative to HTTP method types but IMO CLIs require another level of specificity.

```sh
$ spin cloud apps get "lottery"
Name: lottery
Description: Draw a Magic 8 Ball Winner
URL: https://workshoplottery.fermyon.app
```

src/commands/apps.rs
karthik2804: fixed

karthik2804:
@kate-goldenring I have renamed the command to `info`.

kate-goldenring (approved):
Thanks Karthik!

PR #96, opened 2023-09-22
Split up some large command handlers

itowlson:
This PR does two things (in two separate commits, as all of this is likely to be subjective).  It is purely a refactoring and should not change any existing behaviour.  (Okay, okay.  In the vanishingly implausible cases where an app has an empty URL, the print behaviour changes slightly.  Truth in advertising.)

1. It splits the `apps` and `sqlite` command handlers into subcommand handlers.  This removes two levels of indentation from the actual command handling code, to my eye making it easier to track, and creates some visual separation between the subcommand blocks.
2. It streamlines (in my opinion of course!) the rendering part of the `spin cloud apps info` command by moving some logic into helper functions, and teasing apart the "what do we want to print" from the actual printing.  I feel this makes it easier to read as a "top to bottom" function by reducing the nested ifs/matches.

### DIFF
--- a/src/commands/apps.rs
+++ b/src/commands/apps.rs
@@ -52,68 +52,82 @@ struct CommonArgs {
 impl AppsCommand {
     pub async fn run(self) -> Result<()> {
         match self {
-            AppsCommand::List(cmd) => {
-                let client = create_cloud_client(cmd.common.deployment_env_id.as_deref()).await?;
-                let mut app_list_page = client.list_apps(DEFAULT_APPLIST_PAGE_SIZE, None).await?;
-                if app_list_page.total_items <= 0 {
-                    eprintln!("No applications found");
-                } else {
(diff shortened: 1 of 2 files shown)


karthik2804 (approved):
LGTM! Thanks for this. 

As a follow-up - Do we want to use the same structure of `cmd.run()` for the `variables` subcommand as well? 
https://github.com/fermyon/cloud-plugin/blob/main/src/commands/variables.rs

It does look like the functions there are pretty small as they are today though.

itowlson:
Yes, I didn't feel it was worth it for 4-7 lines per block - the logic is mostly farmed out already, just to free functions rather than subcommand associated functions.  I'm happy to do it though if folks feel it worthwhile.

karthik2804:
That makes sense to me!

Issue #101, opened 2023-09-25
spin deploy error missing base


michelleN:
From spin @ main after adding default for base from fermyon/spin#1803 : 

```
➜  http-rust git:(main) ✗ spin deploy
Warning: You're using a pre-release version of Spin (1.6.0-pre0). This plugin might not be compatible (supported: >=1.3). Continuing anyway.
Error: Cannot read spin.toml manifest from "spin.toml"

Caused by:
    could not load application trigger parameters: missing field `base` at line 14 column 1

Learn more at https://developer.fermyon.com/cloud/faq
```

cc/ @lann

PR #102, opened 2023-09-25
chore(Cargo.toml): bump spin crates

vdice:
Bumps to current HEAD commit of Spin, particularly to get https://github.com/fermyon/spin/commit/d041cc0470c39a341de7b709e7cab5e7d5a011ef

Closes https://github.com/fermyon/cloud-plugin/issues/101

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3370,8 +3370,8 @@ dependencies = [
 
 [[package]]
 name = "outbound-http"
-version = "1.5.0"
-source = "git+https://github.com/fermyon/spin?tag=v1.5.0#ca08dd933de32fe09f4c318fbbc1f04853f2085f"
+version = "1.6.0-pre0"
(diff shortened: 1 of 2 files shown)


itowlson:
This makes me wonder if we should be flagging the change to folks like the containerd shim maintainers.

vdice:
> This makes me wonder if we should be flagging the change to folks like the containerd shim maintainers.

Ah, in case example or template apps are used from main/canary with the 1.5.0-based shim?

itowlson:
The templates still have `base` (at least for now) so there's a good chance they won't come across it.  But yes, generally, because it's plausible that people will try to run 1.6 apps on k8s and get a surprising error.

I wonder if in future we should consider minor versions of the manifest format for stuff like this - at least then a host could say "hey you need to update me" rather than "missing required element."

vdice:
> I wonder if in future we should consider minor versions of the manifest format for stuff like this - at least then a host could say "hey you need to update me" rather than "missing required element."

That seems like a good idea to me.

fibonacci1729 (approved):

PR #106, opened 2023-09-29
Fix database printing to show all of an app's databases

kate-goldenring:
Before fix:
```sh
$ spin cloud sqlite list
+------+-------+----------+
| App  | Label | Database |
+=========================+
| todo | hello | mydb     |
+------+-------+----------+
```

After fix
```sh
$ spin cloud sqlite list
+------+---------+--------------------+
| App  | Label   | Database           |
+=====================================+
| todo | default | brilliant-kangaroo |
|------+---------+--------------------|
| todo | hello   | mydb               |
+------+---------+--------------------+
```

### DIFF
--- a/src/commands/sqlite.rs
+++ b/src/commands/sqlite.rs
@@ -249,12 +249,21 @@ fn print_apps<'a>(links: Vec<Link>, databases_without_links: impl Iterator<Item 
     let mut map = BTreeMap::new();
     for link in &links {
         let app_name = link.app_name();
-        map.entry(app_name)
-            .or_insert_with(|| [link.resource_label.label.as_str(), link.resource.as_str()]);
+        let key = format!(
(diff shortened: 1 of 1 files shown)


fibonacci1729 (approved):

kate-goldenring:
Realized the key should take into account label name too since an app could theoretically point to the same db through a different label. Will update

itowlson:
What would be the impact of removing the row of dashes between lines?  It adds a lot of vertical space, but I can see the value for readability if we anticipate app names being highly variable in length.  Or if we dashed only between apps rather than between every line?

kate-goldenring:
@itowlson I'd be happy with removing the dashes between rows. I also am looking to add json outputting too which i expect would be used for larger blocks of output

PR #109, opened 2023-10-02
Small clean up in app printing

rylev:
Simplifies handling of app printing. The only semantic difference is that if there is a bug and more than one link is returned for the same database, app, label trio, the previous implementation would not print it out, but this will. Such a situation is clearly a bug, and I think the simpler implementation is preferable.

### DIFF
--- a/src/commands/sqlite.rs
+++ b/src/commands/sqlite.rs
@@ -242,7 +242,12 @@ impl ListCommand {
 }
 
 /// Print apps optionally filtering to a specifically supplied app and/or database
-fn print_apps<'a>(links: Vec<Link>, databases_without_links: impl Iterator<Item = &'a Database>) {
+fn print_apps<'a>(
+    mut links: Vec<Link>,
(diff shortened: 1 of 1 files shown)


src/commands/sqlite.rs
```diff
@@ -246,25 +246,14 @@ fn print_apps<'a>(links: Vec<Link>, databases_without_links: impl Iterator<Item
...
-            .map(|(_, [app_name, label, database])| [app_name, label, database]),
-    );
+    let rows = links.iter().map(|link| {
+        [
+            link.app_name(),
+            link.resource_label.label.as_str(),
+            link.resource.as_str(),
+        ]
+    });
+    table.add_rows(rows);
```
kate-goldenring, line 256: Should we sort by app name?

kate-goldenring (approved):
LGTM with one ignorable rec

radu-matei (approved):

PR #112, opened 2023-10-02
Support outputting a list of databases as JSON

kate-goldenring:
Takes a similar approach as `spin templates list` and provides an option of listing databases as JSON

Example output:

```sh
[
  {
    "database": "brilliant-kangaroo",
    "links": [
      {
        "label": "default",
        "app": "todo"
      }
    ]
  },
  {
    "database": "mydb",
    "links": [
      {
        "label": "hello",
        "app": "todo"
      },
      {
        "label": "foo",
        "app": "bar"
      }
    ]
  }
]
```

### DIFF
--- a/src/commands/sqlite.rs
+++ b/src/commands/sqlite.rs
@@ -1,6 +1,7 @@
 use crate::commands::create_cloud_client;
 use crate::commands::link::Link;
 use crate::opts::*;
+use anyhow::bail;
 use anyhow::{Context, Result};
 use clap::{Args, Parser, ValueEnum};
 use cloud::client::Client as CloudClient;
@@ -78,16 +79,17 @@ pub struct ListCommand {
     /// Filter list by a database
     #[clap(short = 'd', long = "database")]
     database: Option<String>,
-    /// Grouping strategy of list
-    #[clap(value_enum, short = 'g', long = "group-by", default_value_t = GroupBy::App)]
-    group_by: GroupBy,
+    /// Grouping strategy of tabular list [default: app]
+    #[clap(value_enum, short = 'g', long = "group-by")]
+    group_by: Option<GroupBy>,
     /// Format of list
     #[clap(value_enum, long = "format", default_value = "table")]
     format: ListFormat,
 }
 
-#[derive(Debug, Clone, Copy, ValueEnum)]
+#[derive(Debug, Clone, Copy, ValueEnum, Default)]
 enum GroupBy {
+    #[default]
     App,
     Database,
 }
@@ -206,6 +208,10 @@ impl ExecuteCommand {
 
 impl ListCommand {
     pub async fn run(self) -> Result<()> {
+        if let (ListFormat::Json, Some(_)) = (&self.format, self.group_by) {
+            bail!("Grouping is not supported with JSON format output")
+        }
+
         let client = create_cloud_client(self.common.deployment_env_id.as_deref()).await?;
         let mut databases = client
             .get_databases(None)
@@ -224,14 +230,27 @@ impl ListCommand {
             }
         }
 
-        if matches!(self.format, ListFormat::Json) {
-            let json_vals: Vec<_> = databases.into_iter().map(json_list_format).collect();
-
-            let json_text = serde_json::to_string_pretty(&json_vals)?;
-            println!("{}", json_text);
-            return Ok(());
+        match self.format {
+            ListFormat::Json => self.print_json(databases),
+            ListFormat::Table => self.print_table(databases),
         }
+    }
 
+    fn print_json(&self, mut databases: Vec<Database>) -> Result<()> {
+        if let Some(app) = &self.app {
+            databases.retain(|d| {
+                d.links
+                    .iter()
+                    .any(|l| l.app_name.as_deref().unwrap_or("UNKNOWN") == app)
+            });
+        }
+        let json_vals: Vec<_> = databases.iter().map(json_list_format).collect();
+        let json_text = serde_json::to_string_pretty(&json_vals)?;
+        println!("{}", json_text);
+        Ok(())
+    }
+
+    fn print_table(&self, databases: Vec<Database>) -> Result<()> {
         let databases_without_links = databases.iter().filter(|db| db.links.is_empty());
 
         let mut links = databases
@@ -250,8 +269,7 @@ impl ListCommand {
                 return Ok(());
             }
         }
-
-        match self.group_by {
+        match self.group_by.unwrap_or_default() {
             GroupBy::App => print_apps(links, databases_without_links),
             GroupBy::Database => print_databases(links, databases_without_links),
         }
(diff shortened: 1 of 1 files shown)


kate-goldenring:
@itowlson I noticed that in `spin templates list`, we [hide the optionality of json output format](https://github.com/fermyon/spin/blob/0306346e554a8ae2968861ee85d4c6fb6b336e8b/src/commands/templates.rs#L448). Is there a reason for that -- and one that is related to whether it should be hidden here too?

itowlson:
@kate-goldenring Just to reduce clutter in the help text.  In that case, the JSON is primarily expected to be consumed by other programs, not be interactive users, so I felt it was not worth making interactive users look at the flag.

src/commands/sqlite.rs
```diff
@@ -214,6 +224,14 @@ impl ListCommand {
             }
         }
 
+        if matches!(self.format, ListFormat::Json) {
```
itowlson, line 227: This seems to ignore the `--app` filter.

src/commands/sqlite.rs
```diff
@@ -214,6 +224,14 @@ impl ListCommand {
...
+            let json_vals: Vec<_> = databases.into_iter().map(json_list_format).collect();
+
+            let json_text = serde_json::to_string_pretty(&json_vals)?;
+            println!("{}", json_text);
+            return Ok(());
+        }
+
```
itowlson, line 234: This function is starting to feel a bit long to me (subjective I know).  How would you feel about breaking from line 227 on into something like

```
match self.format {
  ListFormat::Table => self.print_table(databases),
  ListFormat::Json => self.print_json(databases),
}
```

i.e. move all the print functionality into dedicated format-specific functions?

itowlson (changes requested):
Looks good (style nit aside) but I think it needs to either respect the `--app` filter or (less desirably) be incompatible with it.

src/commands/sqlite.rs
kate-goldenring: I added filtering by app, but it is slightly different than the table format in that it filters to only show databases that are linked to that app but still shows all the links, since it feels strange to misrepresent the links of a database. For example:
```sh
$ spin cloud sqlite list --format json
[
  {
    "database": "brilliant-kangaroo",
    "links": [
      {
        "label": "default",
        "app": "todo"
      }
    ]
  },
  {
    "database": "mydb",
    "links": [
      {
        "label": "hello",
        "app": "bar"
      },
      {
        "label": "foo",
        "app": "todo"
      }
    ]
  }
]
$ spin cloud sqlite list --format json --app "bar"
[
  {
    "database": "mydb",
    "links": [
      {
        "label": "hello",
        "app": "bar"
      },
      {
        "label": "foo",
        "app": "todo"
      }
    ]
  }
]
```

src/commands/sqlite.rs
kate-goldenring: good idea. Moved it out

src/commands/sqlite.rs
itowlson: Yeah it's not quite clear what "app" filtering should mean when this has a strictly hierarchical view grouped on database.  I guess consumers can always get the unfiltered JSON and apply whatever filtering logic fits their use case.

src/commands/sqlite.rs
```diff
@@ -71,18 +72,21 @@ fn disallow_empty(statement: &str) -> anyhow::Result<String> {
 pub struct ListCommand {
     #[clap(flatten)]
     common: CommonArgs,
+    /// Filter list by an app
     #[clap(short = 'a', long = "app")]
     app: Option<String>,
+    /// Filter list by a database
     #[clap(short = 'd', long = "database")]
     database: Option<String>,
+    /// Grouping strategy of list
```
itowlson, line 81: Apologies, I missed the `group_by` option in my previous review.  How should this interact with the JSON representation?

kate-goldenring: I think of this as the "grouping strategy of a list in a table" and dont think it should apply to the json output. Would it be worth updating the comment or renaming it to `table_group_by`. Or we could re-evaluate whether we should include this `group_by` and instead just enable app based output with optional filtering.

itowlson: I don't think it needs renaming, but if it doesn't work with JSON then I think we should have a check up front that says "you can't use grouping with JSON" rather than simply ignoring it - what do you reckon?

kate-goldenring: I updated it to no longer take a default value via the CLI even though a default is assumed when table is selected. this feels a little messy but i updated the command descriptions to hopefully clarify this a little

itowlson (approved):
🎉  Looks good!

src/commands/sqlite.rs
```diff
@@ -232,15 +271,42 @@ impl ListCommand {
                 return Ok(());
             }
         }
-
         match self.group_by {
```
rylev, line 274: nit: to avoid duplication and make it more obvious that `GroupBy::App` is the default can this become `self.group_by.unwrap_or_default()` (where we mark `App` as the default for `GroupBy`)?

src/commands/sqlite.rs
```diff
@@ -214,6 +229,30 @@ impl ListCommand {
...
+        if let Some(app) = &self.app {
+            databases.retain(|d| {
+                d.links
+                    .iter()
+                    .any(|l| l.app_name.as_deref().unwrap_or("UNKNOWN") == app)
+            });
+        }
+        if let Some(db) = &self.database {
+            databases.retain(|d| d.name == *db);
+        }
```
rylev, line 248: Seems we're repeating a lot of the same logic here between json and table printing. Why not delay the formatting decision until we've munged all the data?

src/commands/sqlite.rs
```diff
@@ -232,15 +271,42 @@ impl ListCommand {
...
+
+#[derive(Serialize)]
+struct DatabasesListJson {
+    database: String,
+    links: Vec<SimplifiedResourceLabel>,
+}
+
+/// A ResourceLabel type without app ID for JSON output
+#[derive(Serialize)]
+struct SimplifiedResourceLabel {
```
rylev, line 305: nit: maybe the name should highlight it's used for json output more than it's simplified?
```suggestion
struct ResourceLabelJson {
```

src/commands/sqlite.rs
```diff
@@ -232,15 +271,42 @@ impl ListCommand {
...
+            })
+            .collect(),
+    }
+}
+
+#[derive(Serialize)]
+struct DatabasesListJson {
+    database: String,
+    links: Vec<SimplifiedResourceLabel>,
+}
```
rylev, line 301: Nit: there shouldn't be a need to have owned strings here:
```suggestion
#[derive(Serialize)]
struct DatabasesListJson<'a> {
    database: &'a str,
    links: Vec<SimplifiedResourceLabel<'a>>,
}
```

Such a change would require you also change this:

```rust
fn json_list_format<'a>(database: &'a Database) -> DatabasesListJson<'a> {
```

rylev (commented):
Seems fine though there does seem to be unnecessary code duplication between the json and table formatting code paths.

src/commands/sqlite.rs
kate-goldenring: One is iterating over `Links` while the other is over `Databases` since the tabular form does one row per link while the json form does one array entry per database. If we want to do links as teh output format for json, we could share code but i think it makes more sense in json to represent a database and then nest all its links

src/commands/sqlite.rs
kate-goldenring: Great simplification! Thank you

PR #123, opened 2023-10-05
Enable updating a link with `spin cloud link sqlite`

kate-goldenring:
Currently, to change the database linked to an app with a label, you have to unlink and then link.
```sh
$ spin cloud link sqlite --app boo-todo --database newname default
Error: Database "newdb" is already linked to app "boo-todo" with label "default"
$ spin cloud unlink sqlite --app boo-todo default
Database 'newdb' no longer linked to app boo-todo
$ spin cloud link sqlite --app boo-todo --database newname default
Database 'newname' is now linked to app 'boo-todo' with the label 'default'
```

With change, do not need to unlink to change a link.
```sh
$ spin cloud link sqlite --app boo-todo --database newname default
Database "newdb" is already linked to app "boo-todo" with label "default". Link to database "newname" instead? yes
Database "newname" is now linked to app "boo-todo" with the label "default"
```

### DIFF
--- a/src/commands/link.rs
+++ b/src/commands/link.rs
@@ -72,6 +72,10 @@ impl SqliteLinkCommand {
         let existing_link_for_other_database = other_dbs
             .iter()
             .find_map(|d| find_database_link(d, &self.label));
+        let success_msg = format!(
+            r#"Database "{}" is now linked to app "{}" with the label "{}""#,
+            self.database, self.app, self.label
+        );
         match (existing_link_for_database, existing_link_for_other_database) {
             (Some(link), _) => {
                 anyhow::bail!(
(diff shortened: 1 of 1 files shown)


src/commands/link.rs
```diff
@@ -82,18 +86,34 @@ impl SqliteLinkCommand {
...
                     link.resource_label.label,
+                    self.database,
                 );
+                if dialoguer::Confirm::new()
+                    .with_prompt(prompt)
+                    .default(false)
+                    .interact_opt()?
+                    .unwrap_or_default()
+                {
+                    CloudClient::remove_database_link(&client, &link.resource, link.resource_label)
```
itowlson, line 102: There's no relink in Cloud?  So this still leaves the app in an unlinked state (albeit for a shorter period)?  That seems like something we should address on the Cloud side at some point.

src/commands/link.rs
```diff
@@ -82,18 +86,34 @@ impl SqliteLinkCommand {
                 );
             }
             (_, Some(link)) => {
-                anyhow::bail!(
-                    r#"Database "{}" is already linked to app "{}" with label "{}""#,
+                let prompt = format!(
+                    r#"Database "{}" is already linked to app "{}" with label "{}". Link to database "{}" instead?"#,
```
itowlson, line 90: I found this prompt a bit confusing.  In the PR description you give the example:

```
$ spin cloud link sqlite --app boo-todo --database newname default
Database "newdb" is already linked to app "boo-todo" with label "default".
```

and my reaction was "but... I never told you to do anything to `newdb`.  I told you to link to `newname`".  For me this would be clearer if the message were app-centric rather than database-centric e.g. `App "boo-todo"'s "default" is currently linked to "newdb". Change to link to "newname" instead?`

itowlson (changes requested):
The request changes is for the prompt - the other comment is a flag for the future.

It's out of scope for this PR, but this reminded me that at some point we need to revisit this stuff to make it scripting/CI friendly.  We've built the experience for interactive users but if someone runs it in CI then either it's going to block forever or the interactions will error out because no terminal (I am not sure if `dialoguer` checks for this case).  Like I say, out of scope for this but something to ponder after 0.4 is out maybe.

src/commands/link.rs
kate-goldenring: There is not but i can set up an issue to track it and add a todo here to use a relink api in the future.

src/commands/link.rs
kate-goldenring: Updated to 
```sh
$ spin cloud link sqlite --app boo-todo --database newdb default
App "boo-todo"'s "default" label is currently linked to "newname". Change to link to database "newdb" instead? yes
Database "newdb" is now linked to app "boo-todo" with the label "default"
```

itowlson (approved):

kate-goldenring:
> We've built the experience for interactive users but if someone runs it in CI then either it's going to block forever or the interactions will error out because no terminal (I am not sure if dialoguer checks for this case). Like I say, out of scope for this but something to ponder after 0.4 is out maybe.

@itowlson I naively thought `dialoguer`'s defaults handle this, but i see it does not based on our avoidance of it in spin https://github.com/kate-goldenring/spin/blob/0ed4c8234f16f525b642309ebe086dff4567ebcc/src/commands/plugins.rs#L483

PR #126, opened 2023-10-05
Add `--accept-defaults` flag to avoid terminal interactivity

### LINKED ISSUE #125: Enable programmatic invocation of commands without terminal interaction
The following scenarios require terminal interaction where a flag such as `--accept-defaults` should be provided for CI scenarios:
- `spin cloud deploy` for an app with entries in `sqlite_databases` (you are prompted to choose a DB)
- `spin cloud link sqlite` when updating a link
- `spin cloud sqlite delete` you are prompted to confirm your choice to delete the database

kate-goldenring:
fixes #125

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -127,6 +127,10 @@ pub struct DeployCommand {
     /// Can be used multiple times.
     #[clap(long = "variable", parse(try_from_str = parse_kv))]
     pub variables: Vec<(String, String)>,
+
+    /// Accept defaults for all prompted terminal interactions
+    #[clap(long = "accept-defaults", takes_value = false)]
+    pub accept_defaults: bool,
 }
 
 impl DeployCommand {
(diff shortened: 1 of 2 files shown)


itowlson:
Are there no cases at the moment where a script might want to run a command and _not_ accept the defaults, e.g. for setting up links during deploy?

kate-goldenring:
> Are there no cases at the moment where a script might want to run a command and not accept the defaults, e.g. for setting up links during deploy?

@itowlson for do you mean to enable them to specify what database to link to? Would you expect something like `--link label1:db1 --link label2:db2`?

itowlson:
For confirmation, you mentioned that we already have `--yes` on database deletion.  Perhaps that should be our general option for non-interactive confirmations?  In this case for the "change existing link" situation.  (If I read it right, the default for "change existing link" is "no", so this is _not_ accepting the defaults.)

Although now I am imagining the case where someone wants to link a label but only if that label _isn't_ already linked.  So they'd want a `--no` flag.  Yuck.  I am not sure how to tackle this.

itowlson:
@kate-goldenring Yeah that is the sort of conundrum I am wondering about - link to existing DBs vs create new, that kind of thing.  I don't think the CI scenarios are so pressing that we need to "unblock" them (although I will probably turn out to be laughably wrong when half our E2E tests complain about it) - maybe we should take a bit more time on this?  Sorry

Issue #124, opened 2023-10-05
More unit test coverage


kate-goldenring:
This repository has very few unit tests. Ideally, adding a command and new api calls would come with a suite of tests. If we create a trait for the cloud [`Client`](https://github.com/fermyon/cloud-plugin/blob/8bb8d74abd6dd662d7363c1ebead5aee199f2ce4/crates/cloud/src/client.rs#L52), we can use [`mockall`](https://docs.rs/mockall/latest/mockall/#async-traits) to mock api calls and ensure our commands are sending and handling data as expected.

kate-goldenring:
@itowlson any thoughts on this approach?

itowlson:
Yes, this sounds like pretty much the only way, although we'll also have to worry about interactions - some of the interactive code is quite entwined with the integration code.  It may be best to do it incrementally to assess how much value it's giving us - my experience of mock/fake-based testing of very integration-heavy code is it can end up as "well we can confirm it makes this series of API calls but 🤷" - hopefully you've had more success with it than I have? - but a few experiments on the hairier bits of the code will give us a sense for that.  Otherwise we might be better off figuring out how to do integration tests in a reliable and not-too-slow way...!

In terms of the technology, I tend to prefer ad hoc fakes.  I've never met a mocking library I didn't hate with the burning fiery passion of a thousand suns, but that may be down to personal taste (or bad luck with libraries).  It's an implementation detail, though, and we can try either approach and change if we're not loving it.

In any event, it would be great to do this - even if we end up struggling with the tests themselves, I'm confident it will drive improvements to the structuring of the code and keep us honest going forward.

PR #127, opened 2023-10-06
Framework for testing interactions with `CloudClient`

kate-goldenring:
This is a proposed approach to increasing our unit tests in this repository by making the `CloudClient` mockable. It uses the `mockall` crate by having the `CloudClient` implement a `CloudClientInterface` that can be mocked. Also has a few tests for examples.

Pros to this approach:
- Can add more tests to ensure user input is erroring as expected
- Can add tests to ensure we are handling the data returned from the cloud 

Cons:
- As @itowlson [mentions](https://github.com/fermyon/cloud-plugin/issues/124#issuecomment-1749708109), with mocks, tests can be verbose and potentially summed up to "we can confirm it makes this series of API calls but 🤷"
- Requires passing the client as an argument to any testable functions

(I'm sure there are other pros and cons i am forgetting)

fixes https://github.com/fermyon/cloud-plugin/issues/124

### DIFF
--- a/crates/cloud/Cargo.toml
+++ b/crates/cloud/Cargo.toml
@@ -6,6 +6,7 @@ edition = { workspace = true }
 
 [dependencies]
 anyhow = "1.0"
+async-trait = "0.1.73"
 cloud-openapi = { workspace = true }
 mime_guess = { version = "2.0" }
 reqwest = { version = "0.11", features = ["stream"] }
--- a/crates/cloud/Cargo.toml
+++ b/crates/cloud/Cargo.toml
@@ -18,3 +18,6 @@ tokio = { version = "1.17", features = ["full"] }
 tokio-util = { version = "0.7.3", features = ["codec"] }
 tracing = { workspace = true }
 uuid = "1"
+
+[features]
+mocks = []
--- a/crates/cloud/Cargo.toml
+++ b/crates/cloud/Cargo.toml
@@ -9,6 +9,7 @@ anyhow = "1.0"
 async-trait = "0.1.73"
 cloud-openapi = { workspace = true }
 mime_guess = { version = "2.0" }
+mockall = "0.11.4"
 reqwest = { version = "0.11", features = ["stream"] }
 semver = "1.0"
 serde = { version = "1.0", features = ["derive"] }
--- a/crates/cloud/src/lib.rs
+++ b/crates/cloud/src/lib.rs
@@ -1,4 +1,5 @@
 pub mod client;
 mod client_interface;
 pub use client_interface::CloudClientInterface;
+#[cfg(feature = "mocks")]
 pub use client_interface::MockCloudClientInterface;
--- a/crates/cloud/src/lib.rs
+++ b/crates/cloud/src/lib.rs
@@ -1,3 +1,4 @@
 pub mod client;
 mod client_interface;
 pub use client_interface::CloudClientInterface;
+pub use client_interface::MockCloudClientInterface;
--- a/crates/cloud/src/lib.rs
+++ b/crates/cloud/src/lib.rs
@@ -1,1 +1,3 @@
 pub mod client;
+mod client_interface;
+pub use client_interface::CloudClientInterface;
--- a/src/commands/link.rs
+++ b/src/commands/link.rs
@@ -1,8 +1,8 @@
 use anyhow::{Context, Result};
 use clap::{Args, Parser};
-use cloud::client::Client as CloudClient;
 use cloud::CloudClientInterface;
 use cloud_openapi::models::{Database, ResourceLabel};
+use uuid::Uuid;
 
 use crate::commands::{client_and_app_id, sqlite::find_database_link};
 use crate::opts::*;
@@ -43,16 +43,19 @@ struct CommonArgs {
 impl LinkCommand {
     pub async fn run(self) -> Result<()> {
         match self {
-            Self::Sqlite(cmd) => cmd.link().await,
+            Self::Sqlite(cmd) => {
+                let (client, app_id) =
+                    client_and_app_id(cmd.common.deployment_env_id.as_deref(), &cmd.app).await?;
+                cmd.link(client, app_id).await
+            }
         }
     }
 }
 
 impl SqliteLinkCommand {
-    async fn link(self) -> Result<()> {
-        let (client, app_id) =
-            client_and_app_id(self.common.deployment_env_id.as_deref(), &self.app).await?;
-        let database = CloudClient::get_databases(&client, None)
+    async fn link(self, client: impl CloudClientInterface, app_id: Uuid) -> Result<()> {
+        let database = client
+            .get_databases(None)
             .await
             .context("could not fetch databases")?
             .into_iter()
@@ -61,7 +64,8 @@ impl SqliteLinkCommand {
             anyhow::bail!(r#"Database "{}" does not exist"#, self.database)
         }
 
-        let databases_for_app = CloudClient::get_databases(&client, Some(app_id))
+        let databases_for_app = client
+            .get_databases(Some(app_id))
             .await
             .context("Problem listing links")?;
         let (this_db, other_dbs): (Vec<&Database>, Vec<&Database>) = databases_for_app
@@ -101,14 +105,16 @@ impl SqliteLinkCommand {
                     .unwrap_or_default()
                 {
                     // TODO: use a relink API to remove any downtime
-                    CloudClient::remove_database_link(&client, &link.resource, link.resource_label)
+                    client
+                        .remove_database_link(&link.resource, link.resource_label)
                         .await?;
                     let resource_label = ResourceLabel {
                         app_id,
                         label: self.label,
                         app_name: None,
                     };
-                    CloudClient::create_database_link(&client, &self.database, resource_label)
+                    client
+                        .create_database_link(&self.database, resource_label)
                         .await?;
                     println!("{success_msg}");
                 } else {
@@ -121,7 +127,9 @@ impl SqliteLinkCommand {
                     label: self.label,
                     app_name: None,
                 };
-                CloudClient::create_database_link(&client, &self.database, resource_label).await?;
+                client
+                    .create_database_link(&self.database, resource_label)
+                    .await?;
                 println!("{success_msg}");
             }
         }
@@ -180,7 +188,7 @@ impl SqliteUnlinkCommand {
                 )
             })?;
 
-        CloudClient::remove_database_link(&client, &database, label).await?;
+        client.remove_database_link(&database, label).await?;
         println!("Database '{database}' no longer linked to app {}", self.app);
         Ok(())
     }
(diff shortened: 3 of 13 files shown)


kate-goldenring:
@itowlson @rylev I'd love your thoughts on this in general is low priority

src/commands/link.rs
```diff
@@ -207,3 +216,100 @@ impl Link {
...
+        let dbs = vec![
+            Database::new("db1".to_string(), vec![]),
+            Database::new("db2".to_string(), vec![]),
+        ];
+        mock.expect_get_databases()
+            .returning(move |_| Ok(dbs.to_owned()));
+
+        let result = command.link(mock, app_id).await;
+        match result {
+            Ok(_) => panic!("Expected error"),
```
rylev, line 243: Nit: this can be achieved with `result.unwrap_err()`

src/commands/link.rs
```diff
@@ -207,3 +216,100 @@ impl Link {
...
+            Err(err) => assert_eq!(
+                err.to_string(),
+                r#"Database "does-not-exist" does not exist"#
```
rylev, line 246: I'm a little worried about doing exact matches on error messages as this can be pretty brittle, but I think it's fine to leave for now and see how it works out in practice.

src/commands/link.rs
```diff
@@ -207,3 +216,100 @@ impl Link {
...
+            label: "label".to_string(),
+            common: Default::default(),
+        };
+        let app_id = Uuid::new_v4();
+        let dbs = vec![
+            Database::new("db1".to_string(), vec![]),
+            Database::new("db2".to_string(), vec![]),
+        ];
+        mock.expect_get_databases()
+            .returning(move |_| Ok(dbs.to_owned()));
```
rylev, line 267:
```suggestion
            .returning(move |_| Ok(dbs));
```

src/commands/link.rs
```diff
@@ -207,3 +216,100 @@ impl Link {
...
```
rylev, line 239:
```suggestion
            .returning(move |_| Ok(dbs));
```

src/commands/link.rs
```diff
@@ -207,3 +216,100 @@ impl Link {
...
+                vec![ResourceLabel {
+                    app_id,
+                    label: command.label.clone(),
+                    app_name: Some("app".to_string()),
+                }],
+            ),
+            Database::new("db2".to_string(), vec![]),
+        ];
+        mock.expect_get_databases()
+            .returning(move |_| Ok(dbs.to_owned()));
```
rylev, line 301:
```suggestion
            .returning(move |_| Ok(dbs));
```

src/commands/sqlite.rs
```diff
@@ -437,3 +448,79 @@ pub fn find_database_link(db: &Database, label: &str) -> Option<Link> {
...
+        let command = CreateCommand {
+            name: "db1".to_string(),
+            common: Default::default(),
+        };
+        let dbs = vec![
+            Database::new("db1".to_string(), vec![]),
+            Database::new("db2".to_string(), vec![]),
+        ];
+        mock.expect_get_databases()
+            .returning(move |_| Ok(dbs.to_owned()));
```
rylev, line 469:
```suggestion
            .returning(move |_| Ok(dbs));
```

rylev (approved):
Thanks for working on this! I think the direction looks good. I've not worked with the `mockall` crate before, but it looks nice.

src/commands/link.rs
kate-goldenring: this has to clone the list because the function could be called more than once and as a non-copy type, it cannot be moved out of multiple times. Ideally, mockall would have a `expect_get_databases_once` function that could take in an `FnOnce`.

kate-goldenring: It already exists!
```rust
mock.expect_get_databases()
            .return_once(move |_| Ok(dbs));
```

src/commands/sqlite.rs
```diff
@@ -437,3 +448,77 @@ pub fn find_database_link(db: &Database, label: &str) -> Option<Link> {
...
+
+    #[tokio::test]
+    async fn test_create_success() -> Result<()> {
+        let mut mock = MockCloudClientInterface::new();
+        let command = CreateCommand {
+            name: "db1".to_string(),
+            common: Default::default(),
+        };
+        let dbs = vec![Database::new("db2".to_string(), vec![])];
+        mock.expect_get_databases().return_once(move |_| Ok(dbs));
```
itowlson, line 486: This layout where the API calls are grouped together was more readable for me.  It might still be good to have some blank lines to break this up into stanzas (the command, the mocked API, the action, that kind of thing).

itowlson (commented):
This is great to see.  I left a few comments but nothing showstopping.

I did, as expected, find the mock-API style of testing a bit tricky.  The lack of asserts at the end of many of the tests threw me - but of course the assertion is actually in setting up the expectation!  I think this is part of why I prefer fakes, so that I can make explicit assertions about before and after state rather than setting expectations about specific actions.  However setting up fakes for something like this would be a _much_ greater amount of work, and this seems like a pragmatic way to get some unit testing going.  Thanks for driving it!

crates/cloud/src/lib.rs
```diff
@@ -1 +1,4 @@
 pub mod client;
+mod client_interface;
+pub use client_interface::CloudClientInterface;
+pub use client_interface::MockCloudClientInterface;
```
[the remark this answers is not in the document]
kate-goldenring: looks like code compiled with `cfg(test)` is [not accessible to external crates](https://github.com/rust-lang/cargo/issues/8379), so I wasn't able to reference the mock type after putting it behind this test flag. But I was able to put it behind a feature flag and ill push up a change for that

src/commands/link.rs
```diff
@@ -207,3 +214,91 @@ impl Link {
...
+            Database::new("db1".to_string(), vec![]),
+            Database::new("db2".to_string(), vec![]),
+        ];
+        mock.expect_get_databases().return_once(move |_| Ok(dbs));
+        let expected_resource_label = ResourceLabel {
+            app_id,
+            label: command.label.clone(),
+            app_name: None,
+        };
+        mock.expect_create_database_link()
```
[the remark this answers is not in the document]
kate-goldenring: No it is not opinionated. Happy to put the expectations next to each other

src/commands/sqlite.rs
kate-goldenring: Good call out -- I updated the grouping to hopefully be more readable

src/commands/link.rs
```diff
@@ -207,3 +214,91 @@ impl Link {
...
+        let result = command.link(mock, app_id).await;
+        assert_eq!(
+            result.unwrap_err().to_string(),
+            r#"Database "does-not-exist" does not exist"#
+        );
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_sqlite_link_success() -> Result<()> {
```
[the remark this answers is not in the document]
kate-goldenring: I made the `sqlite` test names more explicit and readable. Let me know what you think.

crates/cloud/Cargo.toml
```diff
@@ -6,8 +6,10 @@ edition = { workspace = true }
 
 [dependencies]
 anyhow = "1.0"
+async-trait = "0.1.73"
 cloud-openapi = { workspace = true }
 mime_guess = { version = "2.0" }
+mockall = "0.11.4"
```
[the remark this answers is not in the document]
kate-goldenring: It does because it is in a separate crate. I put it behind a feature flag now

src/commands/link.rs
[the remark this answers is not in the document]
itowlson: Thanks!  The `link` test names are definitely more expressive!

src/commands/sqlite.rs
```diff
@@ -437,3 +448,83 @@ pub fn find_database_link(db: &Database, label: &str) -> Option<Link> {
...
     db.links.iter().any(|l| l == link)
 }
+
+#[cfg(test)]
+mod sqlite_tests {
+    use super::*;
+    use cloud::MockCloudClientInterface;
+
+    #[tokio::test]
+    async fn test_create_error_already_exists() -> Result<()> {
```
itowlson, line 458:
```suggestion
    async fn test_create_if_db_already_exists_then_error() -> Result<()> {
```

src/commands/sqlite.rs
```diff
@@ -437,3 +448,83 @@ pub fn find_database_link(db: &Database, label: &str) -> Option<Link> {
...
+        let result = command.run(mock).await;
+        assert_eq!(
+            result.unwrap_err().to_string(),
+            r#"Database "db1" already exists"#
+        );
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_create_success() -> Result<()> {
```
itowlson, line 480:
```suggestion
    async fn test_create_if_db_does_not_exist_db_is_created() -> Result<()> {
```

(to signal that the important thing here is not so much that the command returns Ok(()) but that `create_database()` gets called)

src/commands/sqlite.rs
```diff
@@ -437,3 +448,83 @@ pub fn find_database_link(db: &Database, label: &str) -> Option<Link> {
...
+        mock.expect_get_databases().return_once(move |_| Ok(dbs));
+        mock.expect_create_database()
+            .withf(move |db, rl| db == "db1" && rl.is_none())
+            .returning(|_, _| Ok(()));
+
+        command.run(mock).await
+    }
+
+    #[tokio::test]
+    async fn test_delete_error_does_not_exist() -> Result<()> {
```
itowlson, line 497:
```suggestion
    async fn test_delete_if_db_does_not_exist_then_error() -> Result<()> {
```

src/commands/sqlite.rs
```diff
@@ -437,3 +448,83 @@ pub fn find_database_link(db: &Database, label: &str) -> Option<Link> {
...
+        let result = command.run(mock).await;
+        assert_eq!(
+            result.unwrap_err().to_string(),
+            r#"No database found with name "db1""#
+        );
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_delete_success() -> Result<()> {
```
itowlson, line 516:
```suggestion
    async fn test_delete_if_db_exists_then_it_is_deleted() -> Result<()> {
```

itowlson (approved):
I suggested some names for the create/delete tests - feel free to use or biff them as you prefer!  LGTM regardless - thanks!

crates/cloud/Cargo.toml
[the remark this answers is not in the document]
rylev: Hmm... you should be able to put this behind a `dev-dependency` instead of a feature. Are you running into issues with that?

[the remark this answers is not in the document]
kate-goldenring: Yeah. I cannot seem to figure out how to put it behind a dev dependency and then use it in a testing scope from another crate that imports the crate. Can you import types that are behind a dev dependency from another crate?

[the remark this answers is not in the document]
kate-goldenring: Seems i am running into https://github.com/rust-lang/cargo/issues/8379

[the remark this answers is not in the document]
kate-goldenring: @rylev I am merging this but happy to do a follow up pr if you have ideas about how to put it behind a dev dep

PR #129, opened 2023-10-12
`spin cloud sqlite execute` by label

### LINKED ISSUE #70: Sqlite execute does not respect (default) database
When running sqlite commands the name of the database is required.  This makes it difficult to run ci deployments using migrations without using awk.

```
❯ spin cloud sqlite list
Databases (1)
blissful-sheep (default)

❯ spin cloud sqlite execute default @highscore/migration.sql
Error: No database found with name "default"

❯ spin cloud sqlite execute blissful-sheep @highscore/migration.sql
```

The work around is 
```
❯ spin cloud sqlite execute "$(spin cloud sqlite list | awk '/default/{print $1}')" @highscore/migration.sql
```

itowlson:
Fixes #70.

I noticed there was an existing `database_has_link` function but it needs a UUID.  Should I be fetching one in the app case?  I couldn't see how it would help (because the UUID derives from the name anyway), but am probably missing something!

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -716,7 +716,7 @@ async fn get_database_selection_for_existing_app(
     let databases = client.get_databases(None).await?;
     if databases
         .iter()
-        .any(|d| database_has_link(d, resource_label))
+        .any(|d| database_has_link(d, &resource_label.label, resource_label.app_name.as_deref()))
     {
         return Ok(ExistingAppDatabaseSelection::AlreadyLinked);
     }
--- a/src/commands/sqlite.rs
+++ b/src/commands/sqlite.rs
@@ -54,8 +54,8 @@ pub struct DeleteCommand {
 #[derive(Parser, Debug)]
 pub struct ExecuteCommand {
     /// Name of database to execute against
-    #[clap(value_parser = clap::builder::ValueParser::new(disallow_empty))]
-    name: String,
+    #[clap(short = 'd', long = "database", value_parser = clap::builder::ValueParser::new(disallow_empty))]
+    database: String,
 
     ///Statement to execute
     #[clap(value_parser = clap::builder::ValueParser::new(disallow_empty))]
@@ -215,8 +215,8 @@ impl ExecuteCommand {
             .get_databases(None)
             .await
             .context("Problem fetching databases")?;
-        if !list.iter().any(|d| d.name == self.name) {
-            anyhow::bail!("No database found with name \"{}\"", self.name);
+        if !list.iter().any(|d| d.name == self.database) {
+            anyhow::bail!("No database found with name \"{}\"", self.database);
         }
         let statement = if let Some(path) = self.statement.strip_prefix('@') {
             std::fs::read_to_string(path)
@@ -225,7 +225,7 @@ impl ExecuteCommand {
             self.statement
         };
         client
-            .execute_sql(self.name, statement)
+            .execute_sql(self.database, statement)
             .await
             .context("Problem executing SQL")?;
         Ok(())
--- a/src/commands/sqlite.rs
+++ b/src/commands/sqlite.rs
@@ -54,8 +54,16 @@ pub struct DeleteCommand {
 #[derive(Parser, Debug)]
 pub struct ExecuteCommand {
     /// Name of database to execute against
-    #[clap(short = 'd', long = "database", value_parser = clap::builder::ValueParser::new(disallow_empty))]
-    database: String,
+    #[clap(name = "DATABASE", short = 'd', long = "database", value_parser = clap::builder::ValueParser::new(disallow_empty), group = "db", required_unless_present = "LABEL")]
+    database: Option<String>,
+
+    /// Label of database to execute against
+    #[clap(name = "LABEL", short = 'l', long = "label", value_parser = clap::builder::ValueParser::new(disallow_empty), group = "db", requires = "APP", required_unless_present = "DATABASE")]
+    label: Option<String>,
+
+    /// App to which label relates
+    #[clap(name = "APP", short = 'a', long = "app", value_parser = clap::builder::ValueParser::new(disallow_empty), requires = "LABEL", conflicts_with = "DATABASE")]
+    app: Option<String>,
 
     ///Statement to execute
     #[clap(value_parser = clap::builder::ValueParser::new(disallow_empty))]
@@ -159,7 +167,10 @@ impl SqliteCommand {
                 let client = create_cloud_client(cmd.common.deployment_env_id.as_deref()).await?;
                 cmd.run(client).await
             }
-            Self::Execute(cmd) => cmd.run().await,
+            Self::Execute(cmd) => {
+                let client = create_cloud_client(cmd.common.deployment_env_id.as_deref()).await?;
+                cmd.run(client).await
+            }
             Self::List(cmd) => cmd.run().await,
             Self::Rename(cmd) => cmd.run().await,
         }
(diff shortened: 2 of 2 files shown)


src/commands/sqlite.rs
```diff
@@ -54,8 +54,16 @@ pub struct DeleteCommand {
 #[derive(Parser, Debug)]
 pub struct ExecuteCommand {
     /// Name of database to execute against
-    #[clap(value_parser = clap::builder::ValueParser::new(disallow_empty))]
-    name: String,
+    #[clap(name = "DATABASE", short = 'd', long = "database", value_parser = clap::builder::ValueParser::new(disallow_empty), group = "db", required_unless_present = "LABEL")]
+    database: Option<String>,
+
+    /// Label of database to execute against
+    #[clap(name = "LABEL", long = "label", value_parser = clap::builder::ValueParser::new(disallow_empty), group = "db", requires = "APP", required_unless_present = "DATABASE")]
```
kate-goldenring, line 61: Should we add a short option here?

```suggestion
    #[clap(name = "LABEL", long = "label", short = 'l', value_parser = clap::builder::ValueParser::new(disallow_empty), group = "db", requires = "APP", required_unless_present = "DATABASE")]
```

src/commands/sqlite.rs
```diff
@@ -54,8 +54,16 @@ pub struct DeleteCommand {
 #[derive(Parser, Debug)]
 pub struct ExecuteCommand {
     /// Name of database to execute against
-    #[clap(value_parser = clap::builder::ValueParser::new(disallow_empty))]
-    name: String,
+    #[clap(name = "DATABASE", short = 'd', long = "database", value_parser = clap::builder::ValueParser::new(disallow_empty), group = "db", required_unless_present = "LABEL")]
```
kate-goldenring, line 57: Is using `group` and `required_...` statements redundant for clap?

src/commands/sqlite.rs
```diff
@@ -209,27 +220,65 @@ impl DeleteCommand {
...
+                .into_iter()
+                .find(|d| has_link(d, label, app))
+                .ok_or_else(|| {
+                    anyhow::anyhow!(r#"No database found with label "{label}" for app "{app}""#)
+                }),
+        }
+    }
+}
+
+fn has_link(database: &Database, label: &str, app: &String) -> bool {
```
kate-goldenring, line 277: On the point of this function that you bring up:

> I noticed there was an existing database_has_link function but it needs a UUID. Should I be fetching one in the app case? I couldn't see how it would help (because the UUID derives from the name anyway), but am probably missing something!

ResourceLabel has app_id (UUID) as well as app_name because the latter is needed when listing links  with a user readable app name(so app name doesn't need to be looked up for every app ID).

I'd prefer to have one `has_link`/ `database_has_link` function. Can we change the `database_has_link` function to take in label and app name (like `has_link`)? All app names are unique so that matching should be sufficient

src/commands/sqlite.rs
itowlson: Yeah, I guess also allow `-a` for app for consistency with `spin cloud sqlite list`?  I don't think `-a` is used for anything else on the `sqlite` commands.

src/commands/sqlite.rs
itowlson: The documentation is confusing, but I gave this a test and it seems that `group` means _at most one_ not _exactly one_.  If I leave `group` out then clap lets me write `spin cloud sqlite execute "blah blah blah"` (and then goes down the "can't happen" branch of database resolution).

itowlson: Uh sorry that was if I removed `required_unless_present` let me try again I have confused myself.

itowlson: Okay so:

* If I omit RUP then it lets me omit all flags which is invalid
* If I omit group, and supply all three flags, then it does fail, but with "The argument '--database <DATABASE>' cannot be used with '--app <APP>'" which is maybe okay?  To me it feels less correct than "cannot be used with label" but maybe app is just as good a perspective?

Anyway I will leave both in for now, let me know if you have an opinion about this because I'm not sure I do.

src/commands/sqlite.rs
itowlson: I've done this - I don't love what it ended up doing to either call site but maybe you know a better way to handle the ever-vexing combination of options and strings.  Otherwise we can just live with it, it's only a couple of places.

src/commands/sqlite.rs
kate-goldenring: With and without group i seem to get the same (expected) error for `spin cloud sqlite execute "blah"`
```
error: The following required arguments were not provided:
    --database <DATABASE>
    --label <LABEL>
```

src/commands/sqlite.rs
```diff
@@ -445,8 +487,11 @@ pub fn find_database_link(db: &Database, label: &str) -> Option<Link> {
     })
 }
 
-pub fn database_has_link(db: &Database, link: &ResourceLabel) -> bool {
-    db.links.iter().any(|l| l == link)
+pub fn database_has_link(database: &Database, label: &str, app: &Option<String>) -> bool {
+    database
+        .links
+        .iter()
+        .any(|l| l.label == label && &l.app_name == app)
```
kate-goldenring, line 494: Instead of taking in an owned string we can deref here
```suggestion
pub fn database_has_link(database: &Database, label: &str, app: Option<&str>) -> bool {
    database
        .links
        .iter()
        .any(|l| l.label == label && l.app_name.as_deref() == app)
}
```

src/commands/sqlite.rs
itowlson: Yes but "all flags missing" isn't the error case that's affected in that case.  If I do `spin cloud sqlite execute -d a -a a -l a blah` then

* With group it prints 

```
error: The argument '--database <DATABASE>' cannot be used with:
    --app <APP>
    --label <LABEL>
```

* Without group it prints

```
error: The argument '--database <DATABASE>' cannot be used with '--app <APP>'
```

I prefer the first but maybe it's subjective

src/commands/sqlite.rs
```diff
@@ -209,27 +220,58 @@ impl DeleteCommand {
...
+impl ExecuteTarget {
+    fn find_in(&self, databases: Vec<Database>) -> anyhow::Result<Database> {
+        match self {
+            Self::Database(database) => databases
+                .into_iter()
+                .find(|d| &d.name == database)
+                .ok_or_else(|| anyhow::anyhow!("No database found with name \"{database}\"")),
+            Self::Label { label, app } => databases
+                .into_iter()
+                .find(|d| database_has_link(d, label, &Some(app.clone())))
```
kate-goldenring, line 269:
```suggestion
.find(|d| database_has_link(d, label, Some(app.as_str())))
```

src/commands/deploy.rs
```diff
@@ -716,7 +716,7 @@ async fn get_database_selection_for_existing_app(
     let databases = client.get_databases(None).await?;
     if databases
         .iter()
-        .any(|d| database_has_link(d, resource_label))
+        .any(|d| database_has_link(d, &resource_label.label, &resource_label.app_name))
```
kate-goldenring, line 719:
```suggestion
        .any(|d| database_has_link(d, &resource_label.label, resource_label.app_name.as_deref()))
```

kate-goldenring (commented):
On suggestion to handle the option wrapping for `database_has_link`

src/commands/sqlite.rs
kate-goldenring: I see. You're right. The first is definitely more explanatory! Group sounds good.

src/commands/sqlite.rs
itowlson: Done, thanks!

itowlson:
@kate-goldenring if no further changes could I petition for an approval please?

PR #130, opened 2023-10-13
Move some cloud functions to an extension

itowlson:
This is a pure refactor - no behaviour changes.  (Unless I messed up.)

While working on some reorganisation of the `deploy` module, I noticed we had some instance methods that didn't use `self` and in fact seemed more concerned with the `CloudClient` than with the `DeployCommand`.  For me it felt more readable to pull them out as extension methods on the `CloudClientInterface`.  (My main interest was in moving code out of `DeployCommand`, which could be done without traitifying.  But this felt nicer.)

I also noticed that `commands::get_app_id_cloud` and `DeployCommand::try_get_app_id_cloud` did the same thing (modulo one `anyhow::Context`), so unified them.

I removed the `_cloud` suffixes which were inherited from earlier Spin code where they had to be distinguished from Platform (`_hippo`) versions of the operations.  The suffix is no longer needed in the cloud-specific plugin.  I also took the opportunity to change the argument order and make some strings into `&str` because reduced the need for repeated cloning.

### DIFF
--- a/crates/cloud/src/client_interface.rs
+++ b/crates/cloud/src/client_interface.rs
@@ -11,7 +11,7 @@ use uuid::Uuid;
 
 #[cfg_attr(feature = "mocks", mockall::automock)]
 #[async_trait]
-pub trait CloudClientInterface {
+pub trait CloudClientInterface: Send + Sync {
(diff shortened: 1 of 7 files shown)


vdice (approved):
Nice improvements and simplifications!

PR #131, opened 2023-10-16
Move deploy helpers to submodules and simplify flow

itowlson:
This is (or at least should be) a pure refactoring.  The key changes are:

* Move blocks of related functions out of the main `deploy.rs` file into their own modules, and make only the required entry points public.  This reduces the size of the main file, and enables readers to identify the 'surface' operations without searching through the ancillary ones.
* Break out some large blocks of code into their own functions.  This reduces the size of the main function, to enables readers to read the function without needing to keep conditions on their mental stack across dozens of lines.  (Although the function calls are still vexingly long; maybe I'm still missing a struct or enum to bring these into focus.  Or maybe it's `cargo fmt`'s fault.  That guy is always causing trouble!)
* Move (certain) I/O operations onto interfaces.  Deployment now uses the CloudClientInterface throughout (which required a small change to unrelated variables code, sorry), and I introduced a new Interactor trait to enable substitution of command line prompting.  The latter is, I am sure, at the wrong level of abstraction right now; the right level would enable both mocking _and_ a scripted-answers implementation for CI/non-interactive deployments.

### DIFF
--- a/crates/cloud/src/cloud_client_extensions.rs
+++ b/crates/cloud/src/cloud_client_extensions.rs
@@ -6,13 +6,28 @@ use crate::CloudClientInterface;
 
 #[async_trait]
 pub trait CloudClientExt {
+    async fn add_revision_ref(&self, storage_id: &str, version: &str) -> anyhow::Result<()>;
+
     async fn get_app_id(&self, app_name: &str) -> Result<Option<Uuid>>;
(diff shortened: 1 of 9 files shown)


itowlson:
Some of this now happens in #151 so I'll base any further efforts in this direction on that

PR #132, opened 2023-10-17
Manifest v2 support

itowlson:
For when https://github.com/fermyon/spin/pull/1780 lands.  The new load path moves all of the LockedApp fiddliness into the Spin loader, so we pretty much don't have to do anything to consume it any more!

The Spin crate references will need to be set back to the Fermyon repo at this point - this points them at @lann's branch just for development.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,13 +36,14 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.3"
+version = "0.8.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c99f64d1e06488f620f932677e24bc6e2897582980441ae90a671415bd7ec2f"
+checksum = "cd7d5a2cecb58716e47d67d5703a249964b14c7be1ec3cad3affc295b2d1c35d"
 dependencies = [
  "cfg-if",
  "once_cell",
  "version_check",
+ "zerocopy",
 ]
 
 [[package]]
@@ -175,9 +176,9 @@ dependencies = [
  "log",
  "parking",
  "polling",
- "rustix 0.37.25",
+ "rustix 0.37.26",
  "slab",
- "socket2 0.4.9",
+ "socket2 0.4.10",
  "waker-fn",
 ]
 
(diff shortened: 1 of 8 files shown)


itowlson:
This seems to be having a problem with the Spin `assets-test` test case.  The app deploys but never goes ready, and Cloud just says "failed to prepare your Spin application."  Other apps with zero-length files work fine.  Investigating!

lann:
Might have a logical conflict with @vdice's OCI changes; it's an old branch!

itowlson:
The problem happens when a (v1, haven't tested with v2 yet) manifest omits the application HTTP `base`.  The "failed to prepare" may be spurious, although earlier it was showing the domain as having a problem.  For now I'll work on the basis that this is a client problem with the readiness detection.

itowlson:
Something amiss with the digest computation.

When it has `base`, the digest is `335de18dadb8ac201ef8e4a6252600948166d3bde7493de2abc4ac7b06e618bc`

When it _doesn't_ have `base`, the cloud plugin looks for digest `125f0c859c1a8b8ef8afd97ff1dcfaa24e400c5e6c8f736e239dcf177ffd6e31` but the cloud reports digest `335de18dadb8ac201ef8e4a6252600948166d3bde7493de2abc4ac7b06e618bc`.

The digest is coming from within the `spin_oci` `push_locked` method.  Next step is to try a `spin registry push` and see if the digest lines up with what the registry calculates.

itowlson:
`spin registry push` (to ghcr) comes up with the same value whether it has `base` or `not`.

itowlson:
Correction.  `spin registry push` _does_ come up with different digests depending on whether or not `base` was present, and these match the digests that the cloud plugin looks for.

lann:
We should probably rethink how we track readiness. Cloud already assigns a unique ID to each deployment; we should be able to expose that in the `.well-known` info and use that instead.

itowlson:
Sure.  That might need a Cloud API update - looking at the Swagger it seems like the only place the deployment ID is currently returned is in logs, though I could well be missing something!  @bacongobbler is the deployment ID available during the deployment cycle itself?

bacongobbler:
Yes, it should be. Cloud assigns a unique ID to each deployment and passes that ID to the runtime.

itowlson:
The diagnosis of OCI digest weirdness was **wrong**.  Baseless applications don't come ready _at all_ - the supposed digest issue was down to the old application still being live.

itowlson:
I think this is good to go.  Thank you @lann for the HTTP base fix!

kate-goldenring (approved):

PR #133, opened 2023-10-17
chore(Cargo.toml): bump spin crates to v1.6 branch revision

vdice:
Bump spin crates to https://github.com/fermyon/spin/commit/54c4d83987f0519ef885b532d577845443d4d61b to get fix for https://github.com/fermyon/spin/issues/1787 and related commits.

Full diff: https://github.com/fermyon/spin/compare/0306346...54c4d83

PR #134, opened 2023-10-17
Update to OCI crate that always inlines

itowlson:
I'm not sure how to test this @vdice - was your test just to deploy the Fermyon web site?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,9 +67,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.1"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


vdice:
Oh thank you @itowlson, I forgot about the override still in this repo.  Note that I hope to bump the crates to a more recent commit soon after (https://github.com/fermyon/cloud-plugin/pull/133, which brings in the archive layers fix applicable to apps with large #s of static assets eg the Fermyon website).

For testing inlined content, any Spin app with one or more empty or sub-128byte files is a good test.  Just want to ensure they deploy successfully (and of course the files are served as expected, assuming they are used in tandem with a fileserver).  One classic example is https://github.com/mikkelhegn/docusaurus-example/tree/no_data_blob which was the app that caught the bug in the first place with its empty `.nojekyll` file.

vdice (approved):

itowlson:
Oh, duh, right, this is the 0-byte fix not the 500-layer fix... Thanks, I'll give those a go!

PR #135, opened 2023-10-20
demote healthz to a warning

bacongobbler:
Some commands may still be able to be executed while the service is unhealthy. For example, `spin cloud sqlite list` may still work if the deployment service is temporarily unavailable.


itowlson (approved):

kate-goldenring (approved):
Is there a separate endpoint we can use or create instead, @bacongobbler?

PR #143, opened 2023-10-31
Break spurious dependency on wasmtime

itowlson:
look at all that beautiful red ink in Cargo.lock just look at it


fibonacci1729 (approved):
🥳

PR #147, opened 2023-11-03
feat(deploy.rs): insert cloud plugin version into locked app metadata

vdice:


src/commands/deploy.rs
```diff
@@ -328,6 +328,11 @@ impl DeployCommand {
             }
         };
 
+        // Insert cloud plugin version into locked app metadata
```
itowlson, line 331: It's a nit, but my preference here is to follow the `ensure_http_base_set` pattern (line 349).  This:

* moves the fiddling to after validation
* localises the mutability to the thing doing the mutation
* declares intent via the function name

src/commands/deploy.rs
```diff
@@ -328,6 +328,11 @@ impl DeployCommand {
...
+        locked_app
+            .metadata
+            .insert("deploy_agent".to_owned(), crate::VERSION.into());
```
itowlson, line 334: This presumes that cloud-plugin is the only deployment agent we ever need to think about.  I'd suggest something structured like:

```
"deployed_by": {
    "agent": "fermyon/cloud-plugin",
    "version": (whatever)
}
```

Otherwise I would call the field something like `cloud_plugin_version` and then other agents could have separate fields.

itowlson (changes requested):
The style thing is only a nit but I would strongly prefer to tweak the proposed metadata because that we are going to be stuck with...

itowlson:
I guess I am also a bit concerned about unconditionally sticking this into the lockfile because it means that what gets deployed to the cloud is not what the user has in their registry.  I know we do it for the HTTP base but that was just a patch until Cloud could handle baseless apps; this seems like a forever thing.  Maybe in future we'll be able to put it out of band somewhere.  Or maybe I'm being too precious about preserving identity.  I wish we had longer to think about this though.

itowlson:
(that last comment is not a thing to worry about changing at this point, just me thinking out loud)

lann (approved):

vdice:
@itowlson: feedback addressed in https://github.com/fermyon/cloud-plugin/pull/147/commits/6398eee761904920d2eeaa15ff44ecf839a9fe63; merging.

PR #149, opened 2023-11-03
chore(*): bump version to 0.5.1

vdice:
Diff is: https://github.com/fermyon/cloud-plugin/compare/v0.5.0...main

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -460,7 +460,7 @@ dependencies = [
 
 [[package]]
 name = "cloud"
-version = "0.5.0"
+version = "0.5.1"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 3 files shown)


adamreese:
Does `https://github.com/fermyon/cloud-plugin/blob/main/spin-pluginify.toml#L3` or `https://github.com/fermyon/cloud-plugin/blob/main/.github/workflows/audits.yml#L12` need to get bumped also?

vdice:
Thanks for double-checking @adamreese.  We are bumping the former (spin-pluginify.toml) but the latter is the cargo vet version (happens to equal the current 0.5.0 cloud plugin version)

adamreese:
👍 Just thought I'd check

adamreese (approved):

PR #154, opened 2023-11-14
hide environment-name argument

### LINKED ISSUE #146: Hide the `--environment-name` option
The `--environment-name` option originally existed for `spin deploy`, to allow users to choose between Fermyon Platform instances.  In the cloud plugin, the only instance of the cloud is, er, the cloud, and the only case for custom environments is for folks within Fermyon working on the cloud.  Hiding the option would get a wordy and obscure option out of users' faces; we could retain it for dev-test purposes but document it in the repo rather than in the CLI.

karthik2804:
closes #146

### DIFF
--- a/src/commands/apps.rs
+++ b/src/commands/apps.rs
@@ -45,7 +45,8 @@ struct CommonArgs {
     #[clap(
         name = "environment-name",
         long = "environment-name",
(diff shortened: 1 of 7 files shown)


itowlson (approved):
image

PR #141, opened 2023-10-26
`spin cloud deploy`: Drop calls to create/update channel

bacongobbler:
New updates to the Fermyon Cloud API made the calls to `POST /api/channels` and `PATCH /api/channels` redundant as the active revision is calculated based on the app's list of revisions. A client now only has to call `create_app` followed by `add_revision` to trigger a deploy to production. On follow-up deploys, only a call to `add_revision` is necessary.

PR #158, opened 2023-12-01
add more context to certain errors

bacongobbler:
This allows the Spin CLI to receive more contextual errors which helps with debugging.

Old error format:

```
><> spin deploy
Uploading spin-checklist version 0.1.0 to Fermyon Cloud...
Deploying...
Error: response status code: 404 Not Found
```

New error format:

```
><> spin deploy
Uploading spin-checklist version 0.1.0 to Fermyon Cloud...
Deploying...
Error: Unable to add revision: response status code: 404 Not Found
```

### DIFF
--- a/src/commands/deploy.rs
+++ b/src/commands/deploy.rs
@@ -253,7 +253,8 @@ impl DeployCommand {
 
                 client
                     .add_revision(storage_id.clone(), version.clone())
-                    .await?;
+                    .await
+                    .context(format!("Unable to upload {}", version.clone()))?;
 
                 for kv in self.key_values {
                     client


src/commands/deploy.rs
```diff
@@ -223,10 +223,12 @@ impl DeployCommand {
...
                 client
                     .add_revision(storage_id.clone(), version.clone())
-                    .await?;
+                    .await
+                    .context("Unable to add revision")?;
                 let existing_channel_id = client
                     .get_channel_id(app_id, SPIN_DEPLOY_CHANNEL_NAME)
-                    .await?;
+                    .await
+                    .context("Could not get existing channel id")?;
```
itowlson, line 231: Revisions and channel IDs are extremely internal concepts.  It does give us a better string if the user reports the error, but I wonder if we could make it clearer that this is "you are not expected to understand this" terminology.  "Internal error" might work but I'm guessing these would also be displayed for network hiccups which are definitely not internal... not sure what to do here.

(And I admit we are already prone to presenting internal terminology to the user anyway...)

bacongobbler: We could change the terminology for revisions to suggest "Could not upload my-app v1.0.0".

The concept of a channel is going away in #141, so I could wait until that drops and rebase this PR against it. Then we don't have to worry about how we want to communicate that piece of terminology to the Spin user.

itowlson: "Could not upload" sounds fine as long as we remember that it doesn't necessarily refer to the 'upload the bits' but to any part of the process.  (Another idea: something like "could not register app with Fermyon Cloud".)  We could do the same text for both revisions and channels for now, with the latter going away at #141 - unless #141 is imminent in which case yep we could hold off, whatever you feel is best!

bacongobbler: Rebased and changed the error message to "Unable to upload v1.0.0".

itowlson (approved):

PR #173, opened 2024-01-22
chore(links): rename database module to resource

kate-goldenring:
Follow up to https://github.com/fermyon/cloud-plugin/pull/170


rylev (approved):
Looks good! I believe the `resource.rs` file still has a TODO in it to be renamed which can now be removed. You might want to consider adding some `//!` module docs to that module briefly describing what a resource is so that those unfamiliar with the nomenclature can be brought up to speed.

PR #175, opened 2024-01-25
fix(kv): shorten KV link store identifier to just 'store'

kate-goldenring:


src/commands/link.rs
```diff
@@ -39,8 +39,8 @@ pub struct KeyValueStoreLinkCommand {
     /// The app that will be using the key value store
     app: String,
     /// The key value store that the app will refer to by the label
-    #[clap(short = 's', long = "key-value-store")]
-    key_value_store: String,
+    #[clap(short = 's', long = "store")]
```
itowlson, line 42: I wonder if we should alias both this and the SQLite one to `name`, save the user having to remember the different option name for each command.  `spin link kv --store` has a nice readability to it though!

itowlson (approved):

src/commands/link.rs
kate-goldenring: Maybe we add the alias `--name` to both of them?

itowlson: Thinking more about how this reads, it would actually end up as `spin cloud link kv --app thing-one --name thing-two`, which doesn't make it clear that the `name` refers to the thing being linked.  Let's punt on this maybe?  Sorry for the counterproductive thought!

PR #177, opened 2024-01-29
fix(links): log resource creation at creation time

kate-goldenring:
Nit. Previous print did not display resource name and would potentially say "creating" when was linking to existing.


src/commands/deploy/resource.rs
```diff
@@ -281,12 +280,14 @@ pub(super) async fn create_resources_for_new_app(
             ResourceSelection::New(r) => {
                 match resource_type {
                     ResourceType::Database => {
+                        println!("Creating {resource_type} named '{r}'");
```
itowlson, line 283: Seems like this could be lifted above the `match` expression?

itowlson (approved):
Thanks for catching this!  One possible style tweak but just a thought.

PR #189, opened 2024-03-19
upload to separate artifacts

bacongobbler:
#188 bumped up from upload-artifact v3 to v4, which introduces a significant breaking change: uploads to artifacts are now immutable.

See https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md


vdice (approved):
TIL.  Thanks for the link.

PR #212, opened 2024-11-05
Add support for formatting output of `cloud apps list` and `cloud apps info`

ThorstenHans:
This PR adds the `--format` flag (values `Plain` & `Json`) to the following commands:

- `spin cloud apps list`
- `spin cloud apps info <app_name>`

The enum contains the `Plain` variant (which is also set as default value for both commands) due to backward compatibility reasons.

### DIFF
--- a/src/commands/apps.rs
+++ b/src/commands/apps.rs
@@ -22,8 +22,8 @@ pub struct ListCommand {
     #[clap(flatten)]
     common: CommonArgs,
     /// Desired output format
-    #[clap(value_enum, long = "format")]
-    format: Option<OutputFormat>,
+    #[clap(value_enum, long = "format", default_value = "plain")]
+    format: OutputFormat,
 }
 
 #[derive(Parser, Debug)]
@@ -41,8 +41,8 @@ pub struct InfoCommand {
     #[clap(flatten)]
     common: CommonArgs,
     /// Desired output format
-    #[clap(value_enum, long = "format")]
-    format: Option<OutputFormat>,
+    #[clap(value_enum, long = "format", default_value = "plain")]
+    format: OutputFormat,
 }
 
 impl AppsCommand {
--- a/src/commands/apps.rs
+++ b/src/commands/apps.rs
@@ -1,8 +1,10 @@
-use crate::commands::{client_and_app_id, create_cloud_client, CommonArgs};
+use crate::commands::{apps_output::AppInfo, client_and_app_id, create_cloud_client, CommonArgs};
 use anyhow::{Context, Result};
 use clap::Parser;
 use cloud::{CloudClientInterface, DEFAULT_APPLIST_PAGE_SIZE};
-use cloud_openapi::models::{AppItem, AppItemPage, ValidationStatus};
+use cloud_openapi::models::{AppItem, ValidationStatus};
+
+use super::apps_output::{print_app_info, print_app_list, OutputFormat};
 
 #[derive(Parser, Debug)]
 #[clap(about = "Manage applications deployed to Fermyon Cloud")]
@@ -19,6 +21,9 @@ pub enum AppsCommand {
 pub struct ListCommand {
     #[clap(flatten)]
     common: CommonArgs,
+    /// Desired output format
+    #[clap(value_enum, long = "format")]
+    format: Option<OutputFormat>,
 }
 
 #[derive(Parser, Debug)]
@@ -35,6 +40,9 @@ pub struct InfoCommand {
     pub app: String,
     #[clap(flatten)]
     common: CommonArgs,
+    /// Desired output format
+    #[clap(value_enum, long = "format")]
+    format: Option<OutputFormat>,
 }
 
 impl AppsCommand {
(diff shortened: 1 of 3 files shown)


src/commands/apps_output.rs
```diff
@@ -0,0 +1,92 @@
...
+            };
+        }
+        Ok(())
+    }
+}
+
+pub(crate) fn print_app_list(apps: Vec<String>, format: Option<OutputFormat>) {
+    let info = match format {
+        Some(OutputFormat::Json) => serde_json::to_string_pretty(&apps).unwrap(),
+        _ => apps.join("\n"),
```
itowlson, line 80: Make `format` non-optional and make this an explicit match, so it gets caught if we add a case to `OutputFormat`.

src/commands/apps_output.rs
```diff
@@ -0,0 +1,92 @@
...
+    };
+    println!("{}", info);
+}
+
+pub(crate) fn print_app_info(app: AppInfo, format: Option<OutputFormat>) {
+    match format {
+        Some(OutputFormat::Json) => {
+            println!("{}", serde_json::to_string_pretty(&app).unwrap())
+        }
+        _ => print!("{}", app),
```
itowlson, line 90: Again, prefer an explicit match.

src/commands/apps.rs
```diff
@@ -35,6 +40,9 @@ pub struct InfoCommand {
     pub app: String,
     #[clap(flatten)]
     common: CommonArgs,
+    /// Desired output format
+    #[clap(value_enum, long = "format")]
+    format: Option<OutputFormat>,
```
itowlson, line 45: Use clap `default_value` so that this is not an Option.  (cf. https://github.com/fermyon/spin/blob/d83330147584f87618dcbbb75d6422b660f3c39d/src/commands/templates.rs#L465)

src/commands/apps.rs
```diff
@@ -19,6 +21,9 @@ pub enum AppsCommand {
 pub struct ListCommand {
     #[clap(flatten)]
     common: CommonArgs,
+    /// Desired output format
+    #[clap(value_enum, long = "format")]
+    format: Option<OutputFormat>,
```
itowlson, line 26: Again, use clap `default_value` so that this is not an Option.

src/commands/apps_output.rs
```diff
@@ -0,0 +1,92 @@
...
+        Ok(())
+    }
+}
+
+pub(crate) fn print_app_list(apps: Vec<String>, format: Option<OutputFormat>) {
+    let info = match format {
+        Some(OutputFormat::Json) => serde_json::to_string_pretty(&apps).unwrap(),
+        _ => apps.join("\n"),
+    };
+    println!("{}", info);
```
itowlson, line 82: For the plain text case, please bring back the "No applications found" message if the list is empty.  (But that message needs to be printed to stderr not stdout.)

src/commands/apps_output.rs
```diff
@@ -0,0 +1,92 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "Name: {}", self.name)?;
+        if self
+            .description
+            .as_ref()
+            .is_some_and(|desc| !desc.is_empty())
+        {
+            writeln!(f, "Description: {}", self.description.clone().unwrap())?;
+        }
+        if self.domain_info.domain.is_some() {
```
itowlson, line 66: `if let Some(domain) = self.domain_info.domain.as_ref()` perhaps?

src/commands/apps_output.rs
```diff
@@ -0,0 +1,92 @@
...
+            let domain = self.domain_info.domain.clone().unwrap();
```
itowlson, line 67: You should be fine with a reference rather than a clone here I think.

src/commands/apps_output.rs
```diff
@@ -0,0 +1,92 @@
...
+
+impl Display for AppInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "Name: {}", self.name)?;
+        if self
+            .description
+            .as_ref()
+            .is_some_and(|desc| !desc.is_empty())
+        {
+            writeln!(f, "Description: {}", self.description.clone().unwrap())?;
```
itowlson, line 64: Again consider an `if let Some(...) = ` rather than `is_some` followed by `unwrap`.  `unwrap` is a red flag that forces readers to go "wait why are we panicking here."

src/commands/apps_output.rs
```diff
@@ -0,0 +1,92 @@
...
+    Plain,
+    Json,
+}
+
+#[derive(Serialize)]
+pub(crate) struct AppInfo {
+    name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    description: Option<String>,
+    url: Option<String>,
```
itowlson, line 17: Not sure why description is `skip_serializing_if` None but this always serialises even if None?  Same with `DomainInfo::domain`.  If this is a deliberate design choice then maybe add some comments about the reasoning behind fields that serialise to `null` vs being omitted.

src/commands/apps_output.rs
```diff
@@ -0,0 +1,92 @@
...
+    }
+}
+
+impl Display for AppInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(f, "Name: {}", self.name)?;
+        if self
+            .description
+            .as_ref()
+            .is_some_and(|desc| !desc.is_empty())
```
itowlson, line 62: It seems like the AppInfo constructor guarantees that `description` can never be a `Some` of the empty string - the constructor appears to transform that case to None.  Consider omitting this check - if you leave it in, consider adding a comment that it's belt and braces in case the constructor logic changes.

(How I wish languages had NonEmptyString types!)

src/commands/apps_output.rs
```diff
@@ -0,0 +1,92 @@
...
+impl AppInfo {
+    pub(crate) fn new(
+        name: String,
+        description: Option<String>,
+        domain: Option<String>,
+        domain_validation_finished: bool,
+    ) -> Self {
+        let url = domain.as_ref().map(|d| format!("https://{}", d));
+        let desc: Option<String> = match description {
+            Some(d) => match d.is_empty() {
```
itowlson, line 38: You can use a guard here to avoid the nested match e.g.

```
match description {
  Some(d) if !d.is_empty => Some(d),
  _ => None,
}
```

src/commands/apps_output.rs
```diff
@@ -0,0 +1,92 @@
...
+
+impl AppInfo {
+    pub(crate) fn new(
+        name: String,
+        description: Option<String>,
+        domain: Option<String>,
+        domain_validation_finished: bool,
+    ) -> Self {
+        let url = domain.as_ref().map(|d| format!("https://{}", d));
+        let desc: Option<String> = match description {
```
itowlson, line 37: Is the type annotation needed?  Feels like the compiler should be able to infer it.

itowlson (changes requested):
The functionality looks fine but there are some style things I'd really like to address if possible - sorry

itowlson (approved):
Thanks for this!  Looks good!

PR #217, opened 2024-12-02
Update Spin dependencies to 3.0

itowlson:
Critics might reasonably argue that I should have done this _before_ the 0.10 release.  But I don't think just-prior-to-3.0 snapshot is missing anything significant.  I'm mostly doing this because Dependabot is mad as a wet hen about some 4th-level indirect dependency that it can't rev by itself.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -65,6 +65,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "allocator-api2"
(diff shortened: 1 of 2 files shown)


vdice (approved):

itowlson:
I dunno why I never merged this but it seems entirely superseded by #223

